PR #16, opened 2026-04-08
feat: label-based access control for Docker resources

### LINKED ISSUE #11: Feature Request: Label-based access control for Swarm resources
## Summary

Now that RBAC has been implemented, it would be great if Cetacean could go one
step further by supporting a label-based access control mechanism similar to
what Portainer offers, allowing fine-grained access management for Docker Swarm
resources directly from stack definitions.

## Motivation

The current ACL system relies on name-based regexp matching, which only works
if a strict naming convention is followed across all resources. This approach is
fragile in practice — any deviation from the expected naming pattern silently
breaks access control. A label-based mechanism would remove this dependency on
naming conventions entirely.

A label-based approach would allow teams to define access control directly in
their stack files, making it versionable and automatable — a natural complement
to the newly introduced RBAC system.

## Proposed Implementation

Inspired by [Portainer's access control labels](https://docs.portainer.io/advanced/access-control#resources-deployed-outside-of-portainer), Cetacean could support the
following labels on `service`, `volume`, `config`, `secret` and `network`
resources:

| Label | Access Granted |
|-------|----------------|
| `cetacean.accesscontrol.public` | All Cetacean users have access to the resource. Takes precedence over team/user assignments. |
| `cetacean.accesscontrol.teams=dev,prod` | Write access restricted to teams `dev` and `prod` only. Can be used in conjunction with `cetacean.accesscontrol.users` |
| `cetacean.accesscontrol.users=bob,adam` | Write access restricted to users `bob` and `adam` only. Can be used in conjunction with `cetacean.accesscontrol.teams` |

## Example
```yaml
services:
  myapp:
    image: myapp:latest
    deploy:
      labels:
        cetacean.accesscontrol.teams: "dev,ops"

networks:
  mynetwork:
    labels:
      cetacean.accesscontrol.public: "true"
```

## Benefits

- Access control defined as code, versionable alongside the stack
- Works for resources deployed outside of Cetacean (GitOps, CI/CD pipelines)
- Consistent experience across all Swarm resource types
- Natural extension of the RBAC system to externally managed resources
- No dependency on naming conventions, unlike the current regexp-based ACL approach

Radiergummi:
## Summary

Adds opt-in label-based ACL (`CETACEAN_ACL_LABELS=true`) that lets teams control resource access directly from their stack definitions using `cetacean.acl.read` and `cetacean.acl.write` Docker labels. Uses the same `user:pattern` / `group:pattern` audience syntax as the existing config policy.

**Rationale:** The existing ACL system requires a central policy file managed by a platform team. Labels let individual teams self-serve — they add access rules to their compose files alongside Traefik routes, Shepherd config, and other operational labels. This matches the mental model of tools like Traefik where the deployer controls behavior via labels.

**How it works:**
- Labels are checked before config/provider grants. If an identity matches a label audience, the label determines their permission (labels can narrow config grants).
- If an identity doesn't match any label audience, explicit config/provider grants still apply — but the implicit allow-all default is suppressed for that resource.
- Tasks inherit labels from their parent service. Labels are strictly scoped to the resource they're on (no stack fan-out).
- Within labels, most permissive wins (additive, no deny rules).

**What's included:**
- `ResourceResolver.LabelsOf()` interface + cache implementation
- Label evaluation in `Can()`, `Filter()`, `HasAnyGrant()`
- `CETACEAN_ACL_LABELS` config flag (env var + TOML)
- Integration panel on service detail page for viewing/editing ACL labels
- Backend integration detector (`detectACL`)
- Debug logging for label-based ACL decisions
- 22 label-specific tests including 11 security edge case tests
- Benchmarks for all new hot-path code
- Documentation (authorization guide, configuration, integrations, test protocol)

Closes #11

## Test plan

- [ ] `go test ./internal/acl/` — 72 tests including label evaluation, precedence, task inheritance, security edge cases
- [ ] `go test ./internal/acl/ -bench=.` — benchmarks for Can/Filter/ParseACLLabels
- [ ] `go test ./internal/integrations/` — integration detector
- [ ] `npx tsc -b --noEmit && npm run lint` — frontend type check + lint
- [ ] Manual: deploy a service with `cetacean.acl.read: "group:ops"`, verify only ops can see it
- [ ] Manual: verify `Allow` header on `HEAD /services/{id}` reflects label-based permissions
- [ ] Manual: verify ACL panel appears on service detail page and supports inline editing


Radiergummi:
Superseded by #198, which ports this onto current `main`.

This branch had fallen 209 commits behind with 9 conflicting files, and GitHub could no longer compute its mergeability. Re-applying was cleaner than rebasing 23 commits through that, and closing rather than force-pushing keeps the original work readable here.

Almost all of it survived the port — `internal/acl/labels.go`, the label tests, the detector and the panel merged cleanly. Two defects were found and fixed on the way, both worth recording:

- **The allow-all suppression was global rather than per resource.** Enabling `acl.labels` on a cluster with no policy file made every *unlabelled* resource invisible to everyone. This branch's own `docs/authorization.md` rule 4 and its manual test protocol both describe the opposite, so the code disagreed with its intent; #198 scopes the suppression to resources that actually carry labels.
- **`Filter` was quadratic.** It resolved labels per item, and the real resolver is the cache, which answers a name with a scan — 41 ms and 2000 read-lock acquisitions to filter a page of 2000 services, on every list request, SSE refetch and MCP `find`. The benchmarks here use a map-backed stub, so they could not see it. #198 takes one bulk read per type: 1.2 ms for the same page, and linear.

`#11` stays open until #198 lands, which closes it.

PR #175, opened 2026-09-04
feat(mcp): bound how long a task's result is retained

Radiergummi:
A task-augmented `tools/call` kept its result for the lifetime of the server process unless the client asked for an expiry. `docs/mcp.md` told clients to always send `task.ttl` and listed the leak under "Known limitations" — but guidance is not a bound, and this is the one item on the MCP follow-up list that can degrade a running server rather than merely leaving capability on the table.

## Why it leaked

mcp-go deletes a task record only from `scheduleTaskCleanup`, which it starts **solely** when the client supplied `params.task.ttl`. `MaxConcurrentTasks` caps how many tasks run at once, not how many finished ones are kept — the counter is released when a task completes, its record is not. So an agent mutating services on a schedule accumulated one `CallToolResult` per mutation, for as long as it ran.

## The mechanism

mcp-go offers no server-side default TTL and no exported way into its task map. What it does offer is `AddBeforeCallTool`, which receives a **pointer** to the request it passes to `handleToolCall` on the very next line (`server/request_handler.go:520-521`). Filling the field in there is indistinguishable, to everything downstream, from the client having sent it — so mcp-go's own cleanup does the work and nothing here reimplements retention.

| Setting | Default | Effect |
|---|---|---|
| `CETACEAN_MCP_TASK_TTL` | `15m` | Applied when a call omits `ttl`, or sends `0`/`null` |
| `CETACEAN_MCP_MAX_TASK_TTL` | `1h` | Ceiling on what a call may ask for |

Either is disabled by setting it to zero, which needed a non-negative sibling of `resolveDuration` — zero is a meaning here, not a mistake. The default clears the 5-minute convergence timeout with a collection window to spare, because the TTL runs from **creation**, not completion.

An over-long request is **served with the ceiling, not refused**: the caller asked for a cluster mutation, and failing it over a retention preference would be the wrong trade. It lands in the debug log instead.

## What the tests hold

Two things, and the second is the reason the coverage is where it is:

- `boundTaskTTL` must leave a nil `Params.Task` **nil**. The hook sees every `tools/call`, so inventing task params would make mcp-go answer an ordinary synchronous call with a task handle instead of the tool's result — a silent, total break of the write surface. `TestPlainCallIsNotTurnedIntoATask` guards it.
- The line adjacency above is **not a documented mcp-go contract**. So `TestTaskWithoutTTLIsStillReleased` drives a real `tools/call` through the transport with a millisecond TTL and waits for the record to actually go, rather than asserting on the clamp in isolation. A future bump that reorders or copies the request between those two lines fails the build instead of quietly restoring the leak. Both retention tests were watched failing for that exact reason before the hook existed.

Unit coverage on `boundTaskTTL` itself: absent → default, `0`/`-1` → default (non-positive is no cleanup at all, the same leak as absence), over → ceiling, inside → untouched, nil task → untouched, either setting zeroed → that half off, and a default misconfigured above the ceiling still clamped.

## Verification

- `go test ./...` — full suite passes
- `go test ./internal/mcp/ ./internal/config/ -race -count=1` — clean
- `golangci-lint run ./...` — 0 issues

Docs: `docs/mcp.md` (the "Always send `task.ttl`" section becomes "Task retention", and the matching limitations bullet now describes the real behaviour — a result can expire, so poll before it does), `docs/config.reference.toml`, `CLAUDE.md`, CHANGELOG.

Closes the task-retention item from the MCP follow-up review.


Radiergummi:
Both Copilot findings on this PR are incorrect, and I'm resolving them. Recording the reasoning here since it turns on a language change that's easy to miss.

**`new(...)` takes a value in Go 1.26.** The `newexpr` proposal shipped: `new(expr)` allocates, stores the value, and returns the pointer. This repo is on `go 1.26.6` (`go.mod:3`), so `new(def.Milliseconds())` is valid.

It isn't merely allowed, either — it's what the linter demands. My first draft used a `ptrTo[T any](v T) *T` helper, and `golangci-lint run` failed it:

```
internal/mcp/tasks.go:146:14: newexpr: call of ptrTo(x) can be simplified to new(x) (modernize)
internal/mcp/tasks.go:158:6:  newexpr: ptrTo can be an inlinable wrapper around new(expr) (modernize)
```

Taking the suggested edit would reintroduce exactly that failure. CI agrees: `lint-go`, `test`, and `build-frontend` are all green on this head, which they could not be if the code did not compile.

**The aliasing concern is doubly moot.** Per-iteration loop variables have been the semantics since Go 1.22, so `ttl := ttl` has been unnecessary for two years — and independently, `new(expr)` *copies* its operand into fresh storage, so there is no address of a loop variable to alias in the first place. Probed on this toolchain:

```go
for _, ttl := range []int64{0, -1} { got = append(got, new(ttl)) }
// distinct pointers: true, values: 0 -1
```

Two distinct pointers holding both values, which is what the test asserts.

PR #199, opened 2026-09-08
feat: one public URL for the OAuth issuer, OIDC callback and feed links

Radiergummi:
Cetacean had three unrelated answers to "what URL do clients reach me at?" — `mcp.issuer` for the OAuth issuer,
`auth.oidc.redirect_url` for the sign-in callback, and the `X-Forwarded-*` headers for Atom and JSON Feed links.
Each had to be configured separately, and one of them was derived wrong.

This replaces them with a single `server.public_url`, and fixes the derivation.

## The bug

With MCP enabled behind a reverse proxy and no `mcp.issuer` set, the issuer was derived from the listen address.
The default listen address is `:9000`, which has no host — so the issuer came out as `http://:9000`, an origin no
client can resolve. The OAuth discovery chain advertised it anyway, and sign-in failed somewhere the error did not
point at.

The derivation now lives in `internal/config` where it can be tested, rather than in `main.go` where it could not.
When the listen address yields no reachable host and MCP OAuth is genuinely in use, startup stops and names the
setting to fix, rather than advertising an address that cannot work. Auth mode `none` registers no OAuth endpoints,
so it is not affected.

## The setting

`server.public_url` (`-public-url`, `CETACEAN_PUBLIC_URL`) is the canonical external URL, origin only — a path,
query or fragment is a config error, because the path prefix is `server.base_path`'s job and is already applied to
both inbound and outbound paths. It supplies the default for:

- **`mcp.issuer`** — set it only to give MCP a different URL from the rest of Cetacean.
- **`auth.oidc.redirect_url`** — derived as `{public_url}{base_path}/auth/callback`, so it is no longer required
  when `server.public_url` is set. It still overrides, for a callback that lives elsewhere.
- **Atom and JSON Feed links** — built from `server.public_url` when set, rather than from `X-Forwarded-Host` and
  `X-Forwarded-Proto`, which a client can set. Without it, the existing header behaviour is unchanged.

Precedence throughout is explicit setting > `server.public_url` > existing fallback, so no existing configuration
changes meaning. `server.cors.origins` is deliberately untouched: it lists third-party origins, not Cetacean's own.

## Verification

`make check` is green — golangci-lint, oxlint, gofmt/oxfmt and `go test ./...`. The new behaviour is covered by
tests in `internal/config` (issuer derivation, the origin-only validator, the OIDC redirect default) and
`internal/api` (feed links from `public_url` versus from headers).

## Documentation

`CLAUDE.md` and `CHANGELOG.md` are updated here. The user-facing guides — the `configuration` reference entry, the
MCP guide's callout and the OIDC section's note — are written but land with the docs branch, which they depend on
for its link infrastructure.

https://claude.ai/code/session_01FSFjBrr9pkYpkHagXM1tdB

### DIFF
--- a/internal/api/basepath.go
+++ b/internal/api/basepath.go
@@ -69,9 +69,8 @@ func absURL(r *http.Request, path string) string {
 }
 
 // publicURLMiddleware stores server.public_url in the request context so
-// absURL can build links from configuration rather than from X-Forwarded-*,
-// which nothing validates against server.trusted_proxies — any client can set
-// those headers. A no-op when public_url is unset.
+// absURL builds links from configuration rather than from X-Forwarded-*,
+// which any client can set. A no-op when public_url is unset.
 func publicURLMiddleware(publicURL string, next http.Handler) http.Handler {
 	if publicURL == "" {
 		return next
--- a/internal/api/basepath.go
+++ b/internal/api/basepath.go
@@ -11,10 +11,23 @@ type basePathCtxKey struct{}
 
 var basePathKey = basePathCtxKey{}
 
+type publicURLCtxKey struct{}
+
+var publicURLKey = publicURLCtxKey{}
+
 // BasePathFromContext extracts the base path from the context.
 // Returns "" if not set.
 func BasePathFromContext(ctx context.Context) string {
 	if v, ok := ctx.Value(basePathKey).(string); ok {
+		return v
+	}
+	return ""
+}
+
+// PublicURLFromContext extracts the configured external origin.
+// Returns "" when server.public_url is unset.
+func PublicURLFromContext(ctx context.Context) string {
+	if v, ok := ctx.Value(publicURLKey).(string); ok {
 		return v
 	}
 	return ""
@@ -30,10 +43,15 @@ func absPath(ctx context.Context, path string) string {
 	return base + path
 }
 
-// absURL builds a full absolute URL (scheme://host/base/path) from the request.
-// Uses X-Forwarded-Proto/Host when present (reverse proxy), falling back to
-// r.TLS and r.Host. Intended for Atom feeds where RFC 4287 requires IRIs.
+// absURL builds a full absolute URL (scheme://host/base/path) from the
+// request. Uses server.public_url when configured; otherwise
+// X-Forwarded-Proto/Host, falling back to r.TLS and r.Host. Intended for Atom
+// feeds where RFC 4287 requires IRIs.
 func absURL(r *http.Request, path string) string {
+	if base := PublicURLFromContext(r.Context()); base != "" {
+		return base + absPath(r.Context(), path)
+	}
+
 	scheme := "http"
 	if r.TLS != nil {
 		scheme = "https"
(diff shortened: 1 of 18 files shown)


internal/config/config.go
```diff
@@ -36,10 +36,15 @@ const (
...
-	ListenAddr       string
-	BasePath         string          // CETACEAN_BASE_PATH, default ""
+	DockerHost    string
+	PrometheusURL string
+	ListenAddr    string
+	BasePath      string // CETACEAN_BASE_PATH, default ""
+	// PublicURL is the canonical external origin clients reach this
+	// deployment at, e.g. "https://cetacean.example.com". Origin only: the
+	// external path prefix is BasePath, which absPath already prepends to
+	// outbound links. Empty means every consumer keeps its own fallback.
```
Radiergummi: Prune this comment

internal/config/mcp.go
```diff
@@ -415,3 +416,61 @@ func resolveMCPOpsLevel(file *int) (OperationsLevel, error) {
...
+// MCPIssuerRequired reports whether a working /mcp needs a reachable issuer,
+// as opposed to one that only feeds cosmetic tool-icon URLs (main.go's
+// IconBaseURL). OAuth is genuinely in play whenever the upstream auth mode is
+// not "none" and that mode is not fully covered by mcp.oauth.auth_bypass:
+// a bypassed mode (cert, headers, tailscale) authenticates every /mcp request
+// from the upstream provider's own identity instead of a JWT, which is
+// exactly the deployment docs/mcp.md recommends for mTLS clients — one that
+// never drives the OAuth authorize/token flow an unreachable issuer would
+// break. setupMCP downgrades an unreachable issuer to a warning rather than
+// exiting when this returns false.
```
Radiergummi: Trum this comment heavily

internal/config/mcp.go
```diff
@@ -415,3 +416,61 @@ func resolveMCPOpsLevel(file *int) (OperationsLevel, error) {
...
+// reach. server.listen_addr defaults to ":9000", so the derived issuer is
+// "http://:9000" — a URL with an empty host, which resolveMCPIssuer rejects
+// when an operator types it. A wildcard bind ("0.0.0.0", "::") parses to a
+// host but is equally unreachable.
+//
+// The string is returned either way, because how bad an unreachable issuer is
+// depends on the caller: an OAuth client fetches .well-known documents from
+// it and cannot work at all, while auth mode "none" — or an auth mode fully
+// covered by mcp.oauth.auth_bypass, see MCPIssuerRequired — never depends on
+// that flow and ends up only with unreachable icon URLs.
```
Radiergummi: Trim this comment heavily. Do not describe pre-modification state in code comments.

main.go
```diff
@@ -665,8 +666,9 @@ type mcpDeps struct {
 // cache change listener detaches before the cache itself is torn down.
 //
 // The issuer defaults to the listen address + TLS scheme, which only works
-// when no reverse proxy is in front. Behind a proxy, set CETACEAN_MCP_ISSUER
-// (or [mcp].issuer) to the canonical external URL.
+// when the listen address carries a real host and no reverse proxy is in
+// front. Startup fails when OAuth is genuinely in play and no reachable
+// issuer could be derived; see Config.MCPIssuer and Config.MCPIssuerRequired.
```
Radiergummi: Trim this comment heavily.

internal/api/basepath.go
```diff
@@ -50,6 +68,21 @@ func absURL(r *http.Request, path string) string {
 	return scheme + "://" + host + absPath(r.Context(), path)
 }
 
+// publicURLMiddleware stores server.public_url in the request context so
+// absURL can build links from configuration rather than from X-Forwarded-*,
+// which nothing validates against server.trusted_proxies — any client can set
+// those headers. A no-op when public_url is unset.
```
Radiergummi: Trim this comment

internal/config/auth.go
```diff
@@ -371,3 +373,18 @@ func parseScopes(s string) []string {
...
 	return scopes
 }
+
+// defaultRedirectURL builds the OIDC callback URL from server.public_url. The
+// callback route is fixed at GET /auth/callback (internal/auth/oidc.go), so
+// the value is mechanically derivable and only has to be typed when the
+// callback lives somewhere else.
+//
+// Returns "" when public_url is unset, which leaves the existing "oidc mode
+// requires ..." rejection in place rather than inventing a URL.
```
Radiergummi: Trim this comment

PR #213, opened 2026-09-10
feat(api): HTTP semantics — preconditions, async preferences, and compression

Radiergummi:
Implements the 2026-09-09 HTTP semantics design: RFC 9110 preconditions, RFC 7240 async preferences, and response compression. Built from `docs/plans/2026-09-09-http-semantics-plan.md` — 15 tasks, each individually reviewed, plus a whole-branch review and its fix wave, then a second review pass and the fixes below.

## What changes for users

- **`If-Match` on 29 write endpoints.** Read a resource, send its ETag back on the write, and the change is refused with 412 if the resource moved underneath you. Entirely optional: omit the header and every existing write behaves exactly as before.
- **`Prefer: wait=<seconds>`** lets a service mutation wait for the cluster to actually settle before responding, instead of returning the moment Docker accepts the spec. A honoured wait answers with the **settled** service; a timed-out one answers 202 with the last progress line. `respond-async` asks for the opposite.
- **Responses are compressed** with zstd or gzip above a kilobyte, `Vary: Accept-Encoding` either way. A compressed response's ETag can still be handed straight back as `If-Match`.
- **The dashboard loads ~3× lighter on first visit** — assets precompressed at build time and cached immutably. `index.html` stays `no-cache`, so a new release is picked up immediately.
- **Fixed:** `negotiate` clobbered the CORS `Vary` header.

## Structure

- **A0** — a `Chain` middleware type; the router's ten-deep nesting becomes a flat composition.
- **A1** — `precond` middleware plus 29 representation builders, so a precondition compares against the exact bytes the paired GET would emit. Every wrapper sits *after* its ACL and tier constructors, so nothing reaches a 412 without a write grant. Drift-locked by a conformance test that drives the real router and proves absence as well as presence.
- **A2** — the convergence wait moves to `internal/cluster` so REST and MCP cannot disagree on "settled"; `Prefer` applied to six service mutations, covering all seven rows of the spec's wire table.
- **A3** — content-coding negotiation, compression in the ETag write helpers, a Vite precompression plugin, and manifest-driven asset serving.

## Notes for review

- **The ETag hashes identity bytes and appends a coding suffix** (`"abc123-zstd"`), never the compressed bytes — `etagMatchStrong` strips the suffix before comparing, so a validator obtained under one content negotiation still satisfies a precondition sent under another. Both write helpers are pinned by a table-driven round trip, verified by mutation to fail independently.
- **No `ResponseWriter` is wrapped anywhere.** SSE depends on `w.(http.Flusher)` assertions, so compression happens on body bytes inside the write helpers.
- **Feed links carry only the parameters a feed actually reads.** The search feed echoes its own `q` and is the one feed served uncompressed; every compressed feed reflects nothing the caller chose. Verified in both directions against the real router.
- **`AcceptedServiceResponse` repeats a `service` field** rather than embedding `ServiceResponse` — a workaround for a reproduced SIGSEGV in `goccy/go-json` v0.10.6 on an embedded struct carrying a large nested struct plus a nil `omitempty` slice. Reproduced again against this branch's head. The trigger is documented at the type; re-embedding reintroduces a crash that lands *after* the status line is written.
- **The asset manifest stores bare hex ETags** and the Go handler adds the quotes; the variant files are `.gz`/`.zst` while the manifest keys are `gzip`/`zstd`. Both asymmetries are deliberate — "harmonising" either breaks the pair.
- `klauspost/compress` moves from indirect to direct. Same module and version, zero new SBOM entries.

## Caught by the whole-branch review, worth knowing

The precompression plugin originally hooked `generateBundle` with `enforce: "post"`, but Vite orders its own internal post plugins *after* user ones — so it compressed chunk code from before `vite:build-import-analysis` rewrote `__VITE_PRELOAD__`. Every browser negotiates gzip or zstd, so the shipped dashboard would have thrown `ReferenceError` on its first lazy route load. It now runs at `writeBundle` and reads from `outDir`, and `TestEmbeddedManifestMatchesEmbeddedFiles` asserts every manifest ETag equals the hash of the file on disk — an existence check was what let a content defect through a manifest whose purpose is to describe content.

## Caught by the second review pass

Four defects, each fixed with a regression test verified to fail against the old code:

- **`Prefer: wait=<huge>` skipped the wait entirely.** `time.Duration(seconds) * time.Second` overflowed `int64` before the clamp saw it, leaving a negative duration `min` happily kept — so the request answered immediately and reported `Preference-Applied: wait=-8446744073`. The clamp now applies to the second count, before the multiplication.
- **A request resolving to a directory answered 500.** The existence check opened it happily and `serveAsset` cannot seek a directory handle, so `GET /assets` returned `500 internal error` where the old `http.FileServer` had listed it. It now falls through to the SPA shell, as `/assets/` already did.
- **The pooled gzip writer pinned one response body per writer**, holding a reference to the request-local buffer after being returned to the pool.
- **A precondition could not tell a stale validator from an unreachable daemon.** `pluginRepresentation` collapsed every inspect failure into "no current representation", so a conditional `DELETE /plugins/{name}` answered 412 — "your ETag is stale" — when Docker was merely down. `representationFunc` now returns an error rather than a bool: the `errNoRepresentation` sentinel is the 412, and anything else is reported as itself (503/`ENG001` when the engine is unavailable, 500/`ENG004` otherwise). Documented on `removePlugin` in the spec and in `docs/api.md`.

## Test-quality pass

Five assertions claimed more than they checked, and are gone or fixed:

- `frontend/vite.config.test.ts` asserted that `precompress()` appeared in the plugin array — it could only fail if someone deleted that line, and `TestEmbeddedManifestMatchesEmbeddedFiles` already fails outright when the plugin is dropped, because the real build then produces no manifest. Deleted.
- `TestCompressibleEncodingsCoversTheEnum` skipped any enum member whose `String()` returned `"identity"` — which is exactly what an unhandled member returns, so the drift it existed to catch passed silently. The enum gained an `encodingCount` sentinel; the scan now walks that and checks both the list and `String()`.
- `writeEndpointChecksExcluded` silently skipped an excluded endpoint even if the spec documented an `If-Match` on it, against what its own comment promised. It now fails and asks for a driver row.
- The precompress manifest test asserted that no manifest key started with `"."` — over fixture names the test supplies itself. Replaced with a check that no `.vite/` copy was written; the unrelated `entry` assertion moved to its own case, and a redundant pairwise-distinct digest check went (the three equality assertions above it already imply it).
- `TestSSEStillStreamsUnderCompression` asserted "not 500" and "no `Content-Encoding`" without confirming it reached the SSE handler at all; a fall-through to the SPA shell would have satisfied both. It now guards its premise.

`TestEmbeddedManifestMatchesEmbeddedFiles` also pins the build's compression threshold against the server's, which until now only matching comments connected.

## Security scanning

CodeQL raised `go/reflected-xss` on `internal/api/spa.go` because the two `w.Write(preparedIndex)` calls moved into a shared closure, giving an existing sink a new line number. The traced value is `CETACEAN_BASE_PATH` — operator-set, normalized and validated by `config.ValidateBasePath`, then `html.EscapeString`-escaped into a double-quoted attribute. Dismissed as a false positive, matching alerts 12 and 13 on the same file.

## Known gaps

- `/api` (a 179 KB OpenAPI document) is still uncompressed: it bypasses the ETag helpers. Biggest remaining win.
- `Prefer: respond-async, wait=10` that converges still answers with the accept-time snapshot, since the async path short-circuits the settled re-read.
- A handler echoing request-controlled text into an authenticated body must remember to `disableCompression`. Two instances today, both tested; worth a lint rule if the list grows.
- `internal/api/dispatch.go`'s feed `Link` **headers** still reflect the raw query — a header, not a compressed body, so not the BREACH shape, but worth a look.
- Nothing exercises a real non-zero `wait` expiring; the timed-out row uses `wait=0`, because the honest version would take minutes.
- Pre-existing, untouched: REST and MCP disagree on the operations level required to relabel a node (`nodeTier3` vs MCP's documented tier 2).


frontend/vite.config.test.ts
Radiergummi: drop this test please; we don't test config files.

PR #220, opened 2026-09-10
feat(api): refuse cross-origin writes (net/http.CrossOriginProtection)

Radiergummi:
Closes **B4** in the web-standards ledger.

Every state-changing request now passes the stdlib's Fetch Metadata check before it reaches a handler, as a global chain entry between `cors` and `auth`.

### Why it is not only defence in depth

`SameSite=Lax` covers the cookie-borne credential, so for OIDC this is an extra layer. It is not, for the other three modes: a **Tailscale identity, a client certificate and a trusted-proxy header** are ambient credentials a browser attaches with no cookie involved, and `SameSite` has nothing to say about any of them. For those, this is the CSRF defence.

### One trust decision, one value

The CORS allowlist is mirrored in as the trusted origins — an origin CORS admits must not then be refused here.

A wildcard cannot be mirrored: `AddTrustedOrigin` rejects `*`, and reading it as "trust everyone" would disable a security control through a setting that says nothing about CSRF. It warns at startup and cross-origin **writes** stay refused; reads are unaffected. `TestWildcardCORSDoesNotTrustEveryOrigin` pins that ruling rather than leaving it to arrive as a bug report.

The same reasoning removes OIDC's own logout wrapper. Two `CrossOriginProtection` values hold different trusted origins, so a configured cross-origin dashboard would pass the router's check and be refused by the inner one — the CORS/CSRF disagreement this item exists to prevent, one layer down.

### Refusal shape

`Check` is called directly rather than `Handler`, so the response is an RFC 9457 problem under a new **`CSR001`** (new `CSR` domain — a cross-origin block is not a credentials failure, so `AUT005` would mislead). The stdlib's deny path writes plain text, which would be the one refusal in this API carrying no code; its message also names which branch refused, which is what an operator needs.

### Not bypassed

`/mcp` and the OAuth token endpoints looked like candidates for `AddInsecureBypassPattern`. They don't need it: a browser reaches them cross-origin only if CORS permits, and those origins are mirrored. One that isn't configured previously had its request *executed* while the browser discarded the response — for a token exchange, the code burned for nothing. Refusing is strictly better. Non-browser clients send no Fetch Metadata and never reach the check.

### Verification

Tests drive the assembled router against a real write handler, so a permitted request is visibly served rather than merely not-403. Every rule was verified by mutation: removing the chain entry fails four tests, breaking the mirroring fails `TestCrossSiteWriteFromConfiguredOriginIsAllowed`, and restoring the plain-text 403 fails `TestCrossSiteRefusalIsProblemDetails`.

`go test ./...`, `make lint`, `gofmt`, frontend `tsc`/`vitest`/`oxfmt`, and the docs-site build all pass.


docs/api.md
```diff
@@ -309,6 +310,28 @@ resource.
...
+
+Two kinds of client are never affected. Safe methods — `GET`, `HEAD`, `OPTIONS` — always pass, so a cross-origin read
+behaves exactly as CORS allows. So does any client that sends neither `Sec-Fetch-Site` nor `Origin`, which is every
+non-browser client: `curl`, a script, or an [MCP][mcp] host outside the browser.
+
+A wildcard is the one configuration where the two halves cannot agree. `*` is not an origin, so it cannot be trusted
+for writes: browsers may read the API from anywhere while cross-origin writes are still refused. List the origins
+explicitly if a browser client on another origin has to write. That includes a browser-based [MCP][mcp] client such as
+MCP Inspector, whose every call is a `POST`: name the origin it runs on rather than using `*`.
+
```
Radiergummi: This doesn't belong into the documentation.

docs/configuration.mdx
```diff
@@ -76,6 +76,8 @@ the config file.
 
 <ConfigParam name="server.cors.origins" flag="-cors-origins" env="CETACEAN_CORS_ORIGINS">
   Origins allowed to make cross-origin requests to the API, as a comma-separated list, e.g. `https://admin.example.com`, or `*` for any. When unset, no CORS headers are sent and browsers block cross-origin requests.
+
+  These origins are also the ones [cross-site request forgery protection][csrf] trusts, so a browser client listed here can both read the API and write to it. A wildcard cannot be: `*` is not an origin, so browsers may read the API from anywhere while cross-origin writes are still refused. List the origins explicitly if a browser client on another origin has to write.
```
Radiergummi: Unnecessary, can be dropped (or condensed heavily.)

internal/api/csrf.go
```diff
@@ -0,0 +1,52 @@
...
+//
+// Safe methods pass untouched, as do requests carrying neither Sec-Fetch-Site
+// nor Origin — curl, the MCP transport, and every other non-browser client.
+//
+// The CORS allowlist is mirrored in as trusted origins, because the two
+// configurations describe one trust decision: an origin CORS admits must not
+// then be refused here. A wildcard cannot be mirrored — "*" is not an origin,
+// and reading it as "trust everyone" would disable the protection through a
+// setting that says nothing about CSRF. main.go reports that at startup,
+// beside the setting it is about; here it only means nothing is trusted.
```
Radiergummi: Prune this comment aggressively

main.go
```diff
@@ -418,6 +418,19 @@ func main() {
 	if len(cfg.CORSOrigins) > 0 {
 		corsConfig = &api.CORSConfig{AllowedOrigins: cfg.CORSOrigins}
 		slog.Info("CORS enabled", "origins", cfg.CORSOrigins)
+
+		// One setting decides two things: which origins may read the API, and
+		// which may write to it. A wildcard can only answer the first — "*" is
+		// not an origin, so it cannot be a trusted origin for cross-site
+		// request forgery protection, and reading it as "trust everyone" would
+		// disable that protection through a setting that says nothing about it.
```
Radiergummi: Prune this comment aggressively

docs/api.md
Radiergummi: Section removed in 09a0a6dd. The error table row and the `CSR001` registry entry (which carries `Suggestion: "Call the API from an allowed origin, or add yours to the server.cors.origins setting."`) already tell an operator what to add and where, so the prose only restated them at greater length.

Removing it took the `[cross-origin]` anchor with it, so `docs/configuration.mdx` and the `docs/mcp.md` hardening bullet no longer link to it — the one non-obvious fact, that a wildcard is trusted for reads but not for writes, is stated inline in both places instead.

docs/configuration.mdx
Radiergummi: Condensed to one sentence in 09a0a6dd:

> The same origins are trusted for cross-origin writes; `*` cannot be, so with a wildcard browsers may read the API from any origin while cross-origin writes are refused.

Condensed rather than dropped because with the `docs/api.md` section gone this is the only place the wildcard asymmetry is written down, and it is not inferable from the setting name — it is also what invalidated the old `CETACEAN_CORS_ORIGINS='*'` advice for browser MCP clients. Say the word and it goes entirely.

internal/api/csrf.go
Radiergummi: 16 lines to 8 in 09a0a6dd:

```go
// crossOriginProtection refuses non-safe cross-origin browser requests. Not
// only defence in depth: Tailscale, mTLS and trusted-proxy headers all
// authenticate a browser request that carries no cookie at all, so SameSite
// covers none of them.
//
// The CORS allowlist is mirrored in as trusted origins — an origin CORS admits
// must not then be refused here. A wildcard cannot be ("*" is not an origin);
// main.go warns about that at startup.
```

Kept the two things the code cannot say: why this is more than defence in depth, and why the wildcard is not mirrored. The safe-methods/non-browser paragraph went — that is `CrossOriginProtection.Check`'s own documented behaviour, not ours.

main.go
Radiergummi: 5 lines to 2 in 09a0a6dd:

```go
// A wildcard can only answer half of what this setting decides: "*" is
// not an origin, so it cannot be trusted for cross-origin writes.
```

The `slog.Warn` immediately below already says the rest to the operator.

Radiergummi:
Three commits since the review, two of them touching things no comment asked about — flagging what they are so they are not read as drive-by.

### `542713c7` — cleanup pass

- **`setupMCP` re-derived the wildcard rule inline** (`len(origins) == 1 && origins[0] == "*"`), one screen after `CORSConfig.Wildcard()` was added because two readers each deciding for themselves is how they drift. It asks the method now, and its empty-list branch asks `Enabled()`, so `mcpDeps` carries the `CORSConfig` the router already gets.
- **Seven near-identical tests → `TestCrossOriginProtection`'s rows.** They differed in three header values and a status. Each row keeps the comment saying which rule it pins.
- **`TestCrossSiteRequestToAuthRouteIsRefused` could not fail for the reason its name gave.** The chain entry sits above the mux and `NoneProvider.RegisterRoutes` is empty, so `/auth/logout` was not a route in that fixture and *any* path returned 403 — it would have passed with OIDC's wrapper still in place. It survives as a row.
- Dropped the `routerOption`/`withCORS` indirection (one setting, one caller) and the redirect/cookie assertions `TestLogout_ThroughRegisteredRoute` duplicated from `TestLogout_ClearsSession`.

### `7239d3ec` — two gaps that pass turned up

- **`server.public_url` is now a trusted origin.** `Check` compares `Origin` against `r.Host` only on the fallback path — `Sec-Fetch-Site` absent, i.e. a pre-2023 browser — and behind a proxy that rewrites `Host`, that refuses the dashboard's own writes. The only remedy before was listing your own origin in `server.cors.origins`, which also turns on CORS reflection. `ValidatePublicURL` already constrains the value to exactly an origin.
- **`internal/mcp`'s Origin guard read `"*"` differently.** It treated a `*` anywhere in the list as a wildcard while `Wildcard()` requires it to be the only entry, so `CORS_ORIGINS=*,https://dashboard.example` had the API layer matching literally and the guard admitting everything. It takes the decision as `Options.AllowAnyOrigin` now; matching stays exact, so `internal/mcp` still does not import `internal/api`.

### Not done

- **An origin-policy value in `internal/config`, handed to both `api` and `mcp`.** The deeper version of that second fix. It needs either `internal/api` importing `internal/config` or a new shared package, which is a layering call for its own PR.
- **`config.ValidateCORSOrigins`.** An entry like `example.com` (no scheme) is silently ignored by `cors()`'s literal match and only mentioned by cross-origin protection's startup warning. Worth making loud at load, but it is a new config function for a warning that already reaches the operator.

Every rule re-verified by mutation: removing the chain entry fails four tests, breaking the CORS mirroring fails only the row that pins it, dropping the public-URL trust fails `TestPublicURLIsTrusted`, and restoring the guard's own `"*"` check fails `TestOriginGuard`'s new row. `go test ./...` and `golangci-lint run ./...` clean.

PR #221, opened 2026-09-10
fix(api): close the E-block — doc caching, feed links, and the node-label tier

Radiergummi:
Closes **E2, E3 and E4** in the web-standards ledger. Three XS items, batched because each falls out of work already landed.

> **Stacked on #220** (`feat/csrf-protection`) — it needs the router-test fixture that PR refactored, and duplicating it would repeat a review finding from there. Retarget to `main` once #220 merges.

### E2 — `/api` and `/api/scalar.js` bypassed the write helpers

Both wrote their body with a bare `w.Write`, making them the two endpoints compression and conditional caching never reached — and `/api` serves the largest document this API has. Both bodies are fixed for the life of the process, which is exactly what `writeRawWithPrecomputedETag` exists for. Their long `max-age` survives: the helper only fills `Cache-Control` in when unset.

### E3 — feed `Link` headers still reflected the raw query

Links *inside* a feed body were already sanitised to the parameters that feed reads; the `Link` **headers** were pasted from `r.URL.RawQuery`. `addFeedLinks` now calls the same `feedQuery`, which took a `feedData` and now takes the names directly — the header side runs at registration, before any feed is rendered.

`?q=` is consequently declared twice (on the route, and in `searchFeedData`), so `TestSearchFeedDeclaresItsQueryOnBothSides` fails if either goes. One visible change: the href is now canonical (`url.Values.Encode` sorts) rather than raw order, so it matches the feed's own self link exactly.

### E4 — the two transports disagreed on relabelling a node

REST gated `PATCH /nodes/{id}/labels` at tier 3 while MCP's `update_node_labels` sits at tier 2 — and both `docs/mcp-tools.mdx` and the `update_node` description reason explicitly that it belongs at 2, so a deployment can let an agent relabel a node without granting demote-a-manager powers. REST was the side that was wrong.

Moving it surfaced **a third statement of the same tier** in `allow.go`'s `resourceWriteMethods`, which decides the `Allow` header the dashboard gates its affordances on. Left alone it would have hidden the labels editor for an edit the API now accepts.

The ledger asks for one test driving both transports. `internal/api` and `internal/mcp` deliberately do not import each other, so this is two tests naming one rule and each other, following the existing `TestAttachmentEditorsMatchTheRESTTier`.

### Verification

Every rule verified by mutation. **Two mutations initially passed, and both were the tests' fault** — worth recording, since that is the failure shape the #213 working note warns about:

- the API-doc fixtures sat under `compressionThreshold`, so an uncompressed answer looked correct;
- the `Allow` probe asked the labels sub-path, which sets no `Allow` header at all.

Both were rewritten until they failed. `go test ./...`, `make lint` (0 issues), gofmt and the docs-site build all pass.


docs/api.md
```diff
@@ -590,6 +590,12 @@ passes the per-resource [ACL][authorization] write check.
 > [!NOTE]
 > `GET /swarm/unlock-key` returns a credential, so it is gated at level 3 like the writes beside it.
 
+> [!NOTE]
+> `PATCH /nodes/{id}/labels` sits a level below the other node writes. Relabelling a node changes where
+> tasks may be placed; draining it or demoting it from manager changes whether the swarm stays healthy.
+> A deployment can therefore allow relabelling without also allowing a manager to be demoted — the same
+> split [MCP][mcp] makes between `update_node_labels` and `update_node`.
+
```
Radiergummi: These callouts have zero business here. The information belongs into the endpoint docs.

```suggestion
```

internal/api/allow.go
```diff
@@ -39,8 +39,11 @@ var resourceWriteMethods = map[string][]methodSpec{
 		{"DELETE", config.OpsImpactful},    // remove
 	},
 	"node": {
-		{"PUT", config.OpsImpactful},    // availability, role
-		{"PATCH", config.OpsImpactful},  // labels
+		{"PUT", config.OpsImpactful}, // availability, role
+		// Labels are a placement edit, a tier below the writes that decide
+		// whether the swarm stays healthy — matching the route's own gate
+		// and MCP's update_node_labels.
```
Radiergummi: This comment doesn't belong here

```suggestion
```

internal/api/apidoc.go
```diff
@@ -38,21 +43,26 @@ func HandleAPIDoc(specYAML []byte) http.HandlerFunc {
...
-			w.Write(specJSON) //nolint:errcheck
+			spec.serve(w, r)
 		}
 	}
 }
 
 // HandleScalarJS serves the embedded Scalar API reference JavaScript bundle.
+// At 3.7 MB it is the response that most needs staticBody: the route skips
+// authentication, so compressing it per request would let anyone spend ~68ms
+// of server CPU on a 100-byte GET.
```
Radiergummi: No historic detail in code comments.

```suggestion
```

internal/api/apidoc_caching_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// The API documentation endpoints serve two bodies fixed at startup — the
+// OpenAPI document, the largest response this API has, and the Scalar bundle.
+// Both used to be written with a bare w.Write, so neither was compressed and
+// neither carried a validator, while every other endpoint had both. These
+// tests pin them onto the same helper the rest of the tree uses.
```
Radiergummi: No historic detail in code comments.

```suggestion
```

internal/api/apidoc_caching_test.go
```diff
@@ -0,0 +1,130 @@
...
+	{name: "scalar bundle", path: "/api/scalar.js", contains: "scalar"},
+}
+
+func newAPIDocRouter(t testing.TB) http.Handler {
+	t.Helper()
+
+	// Both documents are padded past compressionThreshold, because the real
+	// ones are far past it — the OpenAPI document is the largest response
+	// this API serves — and a fixture under it would let an uncompressed
+	// answer look correct.
```
Radiergummi: No historic detail in code comments.

```suggestion
```

internal/api/dispatch.go
```diff
@@ -13,6 +13,12 @@ import (
 type feedHandlers struct {
 	atom     http.HandlerFunc
 	jsonFeed http.HandlerFunc
+
+	// queryParams names the query parameters these feeds read beyond the
+	// pagination pair — the same declaration feedData.QueryParams carries,
+	// on the registration side, so the alternate Link headers below can
+	// apply feedQuery. Only /search has one.
```
Radiergummi: This comment doesn't add value.

```suggestion
```

internal/api/dispatch.go
```diff
@@ -127,33 +133,32 @@ func (h *Handlers) aclMatchWrap(
 }
 
 // addFeedLinks sets Link headers advertising feed alternates (RFC 8288).
+//
+// The href carries only the parameters the feed it points at actually reads,
+// through the same feedQuery the feed's own links use. Building it from
+// r.URL.RawQuery instead would advertise an alternate that differs from the
+// self link the feed answers with, and would put arbitrary request text in a
+// response header — a second rule for what a feed link may carry, beside the
+// one feedQuery already states.
```
Radiergummi: This comment doesn't add value.

```suggestion
```

internal/api/etag.go
```diff
@@ -111,7 +111,29 @@ func writeRawWithPrecomputedETag(
...
+}
+
+// writeRawNegotiated is the shared body of the precomputed-ETag writers: it
+// negotiates a coding, answers a matching precondition with 304 before
+// touching the body at all, and otherwise writes whatever encode returns for
+// the negotiated coding.
+//
+// encode is a function rather than bytes because a body fixed for the life of
+// the process can cache its codings, and one that is not must compress per
+// response — see staticBody.
```
Radiergummi: Prune this comment, it's too verbose.

internal/api/feed_handlers.go
```diff
@@ -453,16 +453,26 @@ func parseFeedPagination(r *http.Request) (beforeID uint64, limit int) {
 // beyond these declares it in feedData.QueryParams.
 var feedPaginationParams = []string{"before", "limit"}
 
-// feedQuery returns the subset of r's query this feed's links may carry: the
-// pagination pair every feed reads, plus whatever else data declared.
+// searchFeedParams names the parameter only the search feed reads. It is one
+// variable because two code paths need it — the route registration, for the
+// alternate Link header, and searchFeedData, for the links inside the feed —
+// and two literals would be a drift a test could only catch after the fact.
```
Radiergummi: Prune this comment, it's too verbose.

internal/api/feed_handlers.go
```diff
@@ -453,16 +453,26 @@ func parseFeedPagination(r *http.Request) (beforeID uint64, limit int) {
...
+var searchFeedParams = []string{"q"}
+
+// feedQuery returns the subset of r's query a feed's links may carry: the
+// pagination pair every feed reads, plus whatever else the caller declares.
 //
 // The rest of the raw query is attacker-chosen text, and reflecting it into a
 // compressed feed beside ACL-filtered resource names is the BREACH shape. The
 // set is per-feed rather than global because ?q= is read by handleFeedSearch
 // alone, and echoing it from a compressed feed would rebuild that shape.
```
Radiergummi: Prune this comment, it's too verbose.

internal/api/feed_handlers.go
```diff
@@ -453,16 +453,26 @@ func parseFeedPagination(r *http.Request) (beforeID uint64, limit int) {
...
-func feedQuery(r *http.Request, data feedData) url.Values {
+//
+// It takes the extra names rather than a feedData because the alternate Link
+// headers in dispatch.go apply the same rule from the registration side,
+// where no feed has been rendered yet — one rule, two callers.
```
Radiergummi: Prune this comment, it's too verbose.

internal/config/config.go
```diff
@@ -27,10 +27,12 @@ const (
 
 	// OpsConfiguration allows service definition changes: env, labels,
 	// resources, healthcheck, placement, ports, update-policy, rollback-policy,
-	// log-driver.
+	// log-driver. Node labels sit here too: they are a placement edit, and
+	// requiring OpsImpactful to write the labels an OpsConfiguration placement
+	// constraint matches would gate the two halves of one decision apart.
```
Radiergummi: No. This edit only adds unwarranted detail here.

```suggestion
    // log-driver.
```

internal/api/licenses.go
```diff
@@ -6,19 +6,21 @@ import (
 	"github.com/radiergummi/cetacean/internal/api/sbom"
 )
 
-// The three attribution documents are embedded at build time and never change,
-// so their ETags are hashed once here rather than on every request.
+// The three attribution documents are embedded at build time and never
+// change, so each is hashed once here and compressed at most once per coding
+// rather than on every request. Notices alone is 1.3 MB, on a route that —
+// like the rest of /-/ — skips authentication.
```
Radiergummi: No historic references in comments.

```suggestion
```

internal/api/router.go
```diff
@@ -136,6 +148,12 @@ func NewRouter(cfg RouterConfig) http.Handler {
 	svcTier1 := NewChain(svcACL, tier1)
 	svcTier2 := NewChain(svcACL, tier2)
 	svcTier3 := NewChain(svcACL, tier3)
+	// Node labels sit a tier below the rest of a node's writes: relabelling
+	// is a placement-configuration edit, while availability and role can
+	// destabilise the swarm. MCP splits update_node_labels from update_node
+	// for exactly that reason, and REST has to agree — see
+	// TestPatchNodeLabelsIsAdmittedAtTierTwo.
```
Radiergummi: Way too specific and unasked for here.

```suggestion
```

internal/api/router.go
```diff
@@ -66,6 +66,18 @@ func (h *Handlers) listFeeds(title string, eventType cache.EventType) feedHandle
 	}
 }
 
+// searchFeeds builds feedHandlers for the search endpoint, the one feed that
+// reads a query parameter beyond the pagination pair. It is a constructor
+// like listFeeds and detailFeeds so the declaration sits where every other
+// feed's does, rather than inline at the single route that needs it.
```
Radiergummi: No historic references in comments.

```suggestion
```

internal/api/staticbody.go
```diff
@@ -0,0 +1,65 @@
...
+// option for a body that changes but pure waste for one that cannot. The
+// Scalar bundle is 3.7 MB: gzipping it per request costs ~68ms of CPU and a
+// megabyte of garbage, on a route that skips authentication and so can be
+// asked for it by anyone. The attribution documents behind /-/licenses are
+// the same shape at 1.3 MB.
+//
+// Codings are compressed lazily, so a binary nobody fetches the playground or
+// the licence texts from pays nothing, and retained afterwards. Nothing is
+// precomputed at startup: that would move the cost rather than remove it, and
+// pay it for documents most deployments never request.
```
Radiergummi: No historic references in comments.

```suggestion
```

docs/api.md
Radiergummi: Removed. The tier is already in the operations-level table two lines above (`PATCH /nodes/{id}/labels` | 2), so the callout only restated it.

internal/api/allow.go
Radiergummi: Removed. The entry is back to the one-word trailing marker its siblings use — `{"PATCH", config.OpsConfiguration}, // labels` — so the only diff left in this map is the tier value itself.

internal/api/apidoc.go
Radiergummi: Removed; the doc comment is the one line again.

internal/api/apidoc_caching_test.go
Radiergummi: Removed.

internal/api/apidoc_caching_test.go
Radiergummi: Removed.

internal/api/dispatch.go
Radiergummi: Removed. The field now sits bare beside `atom` and `jsonFeed`, which carry none either.

internal/api/dispatch.go
Radiergummi: Cut to the one-line summary plus the sentence stating what the href carries.

internal/api/etag.go
Radiergummi: Pruned to three lines; the paragraph on why `encode` is a function is gone.

internal/api/feed_handlers.go
Radiergummi: Pruned to one line.

internal/api/feed_handlers.go
Radiergummi: Pruned to two lines. I kept the BREACH sentence — it is the reason the function filters at all, so a later edit widening the set needs it — and dropped the rest.

internal/api/feed_handlers.go
Radiergummi: Removed.

internal/config/config.go
Radiergummi: Reverted to `// log-driver.`. The only remaining edit in this block is `availability/labels` to `availability/role` on `OpsImpactful`, which the tier move makes wrong to leave as it was.

internal/api/licenses.go
Radiergummi: Dropped the added sentence; the comment is the pre-existing one plus the word "compressed".

internal/api/router.go
Radiergummi: Removed.

internal/api/router.go
Radiergummi: Cut to `// searchFeeds builds feedHandlers for the search endpoint.`, matching `listFeeds` and `detailFeeds` beside it.

internal/api/staticbody.go
Radiergummi: Cut the cost arithmetic and the startup-precompute paragraph; two sentences left, one of them the laziness note.

docs/api.md
Radiergummi: Follow-up: the pre-existing `GET /swarm/unlock-key` callout above this one is gone too (8f12e12a). The table states the level for every endpoint, that one included.

Radiergummi (commented):
LGTM

PR #235, opened 2026-09-12
feat(api): describe the SSE streams as an AsyncAPI document

Radiergummi:
Cetacean serves **twenty SSE streams**. OpenAPI described four of them, three only partially — it can say what a URL returns as JSON, and has no way to say that the same URL also streams. This adds an AsyncAPI 3.0 document that does.

## What it adds

`GET /api/asyncapi` serves an AsyncAPI 3.0 description of every stream: the eight resource list URLs and eight detail URLs, the global `/events` stream, both log tails and the metrics stream. It names the events each one sends, the shape of their payloads, and which of three cursor dialects its `id:` belongs to — a monotonic history id on the resource streams, an RFC 3339 timestamp that only moves forward on the log tails, and no id at all on `/metrics`.

Two things the document states that the examples alone get wrong: a `batch` frame's payload is an **array** of envelopes rather than one, and a **replayed** event carries no `resource`, because replay reads the change history, which stores identity and not payload.

Every response now carries a second `service-desc` Link beside the OpenAPI one, distinguished by its `type` parameter, and the API catalogue lists it. Cetacean ships no renderer — paste the document into [AsyncAPI Studio](https://studio.asyncapi.com/).

## How it is kept honest

The document is validated against the official AsyncAPI 3.0.0 meta-schema, and three walks hold it to the running server rather than to a list someone maintains:

- every channel address it declares is driven and must actually open a stream;
- every `GET` route the **router** registers is probed with `Accept: text/event-stream`, and anything that streams must appear in the document — the direction that catches a stream added after the document was written;
- the per-type messages are compared against the `cache.EventType` constants read out of the source, so a ninth resource type cannot be added without the contract following.

Frame-level tests drive the real transport for each dialect: event naming and batching on the resource streams, the aged-out-cursor `sync`, the missing `resource` on replay, and the log tail's unnamed, forward-only frames.

## Review fixes riding along

A review of the whole branch turned up six defects, none of them in the AsyncAPI work — they are in code it sits on top of. They are in their own commit (`79a06e8c`) and can be split out if you would rather they landed separately:

- **A filtered listing's CSV link downloaded everything.** `GET /services?search=web` advertised `</services.csv>` with the query dropped, so following the link — the obvious way to hand a filtered view to somebody who would rather read it in a spreadsheet — returned every service.
- **The discovery documents were refused while Docker was down.** `/.well-known/api-catalog` and `/opensearch.xml` read no cluster state, yet both answered ENG001 until the first sync landed — the moment a client probing an unresponsive server most needs them.
- **Cross-origin protection broke browser-based MCP clients.** `/oauth/token`, `/oauth/revoke`, `/oauth/register` and `POST /mcp` authenticate from the request itself, so there is no ambient credential for a cross-origin page to borrow and nothing for the check to defend. RFC 7591 registration is the case that cannot be worked around, since its whole point is a client nobody can list in advance. Consent at `/oauth/authorize` runs under the session cookie and stays checked.
- **The dashboard could not be installed as an app under OIDC.** A web app manifest is fetched with credentials omitted unless the page says otherwise, and `/manifest.webmanifest` is not exempt from authentication.
- **A recommendation that measured zero read as one that measured nothing.** `Current` and `Configured` were `float64` with `omitempty`, so an idle service's zero CPU vanished from the JSON and the CSV alike. They are pointers now, like `Suggested` already was.
- **An upgrade note, not a code change.** Deciding proxy trust at the edge means a deployment behind a TLS-terminating proxy with no `server.trusted_proxies` now names its own internal address in feed links and identifiers, the OpenSearch templates and the API catalogue. Documented in the changelog and on the setting.

## Also in here

The AsyncAPI handler rebuilt the entire ~18 KB document on every request in order to write one `servers` block. The body is a pure function of scheme, host and base path, so it now renders into the same `staticBody` every other served document uses, behind a single-slot cache. Measured on an M3 Pro: **96.9 µs/365 allocs → 6.7 µs/22** identity, **197.5 µs → 2.6 µs** with zstd. Conditional GETs gain the most, since the endpoint sets `max-age=3600` and therefore serves mostly revalidations.

## Verification

`go build ./...`, `go test ./...`, `golangci-lint run ./...` (0 issues), `-race` on the changed packages, `tsc -b --noEmit`, `oxlint`, `oxfmt --check` — all clean.

https://claude.ai/code/session_01TpeaKthnEpS78QyZ7YLLzM

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -15,6 +15,8 @@ info:
 
     Every response (except `/-/` meta endpoints) includes:
     - `Link: </api>; rel="service-desc"` — pointer to this spec
+    - `Link: </api/asyncapi>; rel="service-desc"; type="application/vnd.aai.asyncapi+json;version=3.0.0"` —
+      the AsyncAPI description of the SSE streams
     - `Link: </api/context.jsonld>; rel="describedby"` — JSON-LD context
     - `Vary: Accept` — content negotiation indicator
 
@@ -774,6 +776,9 @@ paths:
                       - href: "https://cetacean.example.com/api"
                         type: "application/json"
                         title: "OpenAPI description"
+                      - href: "https://cetacean.example.com/api/asyncapi"
+                        type: "application/vnd.aai.asyncapi+json"
+                        title: "AsyncAPI description of the event streams"
                     describedby:
                       - href: "https://cetacean.example.com/api/context.jsonld"
                         type: "application/ld+json"
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -13,6 +13,11 @@ info:
     Detail responses include `@context`, `@id`, and `@type`.
     Error responses follow RFC 9457 (Problem Details).
 
+    Both descriptions are served as JSON and as YAML. `/api` and `/api/asyncapi` negotiate on `Accept`
+    (`application/yaml`, `text/yaml`, or the registered `application/vnd.oai.openapi` and
+    `application/vnd.aai.asyncapi+yaml`), and `/api/openapi.yaml` and `/api/asyncapi.yaml` address the
+    YAML form directly.
+
     Every response (except `/-/` meta endpoints) includes:
     - `Link: </api>; rel="service-desc"; type="application/json"` — pointer to this spec
     - `Link: </api/asyncapi>; rel="service-desc"; type="application/vnd.aai.asyncapi+json;version=3.0.0"` —
@@ -655,8 +660,9 @@ paths:
       tags: [Documentation]
       summary: API documentation
       description: |
-        Content-negotiated. HTML requests get the Scalar API playground;
-        all other requests get this OpenAPI spec as JSON.
+        Content-negotiated. HTML requests get the Scalar API playground, a
+        request naming a YAML type gets this document as YAML, and all others
+        get it as JSON.
       security: []
       responses:
         "200":
@@ -666,6 +672,69 @@ paths:
               schema:
                 type: object
             text/html:
+              schema:
+                type: string
+            application/vnd.oai.openapi:
+              schema:
+                type: string
+
+  /api/openapi.yaml:
+    get:
+      operationId: getApiDocYaml
+      tags: [Documentation]
+      summary: OpenAPI description as YAML
+      description: |
+        This document as YAML, at the address the documentation site publishes
+        it under. The source file is served verbatim, so its comments and key
+        order are intact.
+      security: []
+      responses:
+        "200":
+          description: OpenAPI description
+          content:
+            application/vnd.oai.openapi:
+              schema:
+                type: string
+
+  /api/asyncapi:
+    get:
+      operationId: getAsyncApiDoc
+      tags: [Documentation]
+      summary: AsyncAPI description of the event streams
+      description: |
+        Describes every Server-Sent Events stream this server serves — what
+        each channel sends, the shape of its payloads, and which cursor dialect
+        its `id:` belongs to. A request naming a YAML type gets YAML; everything
+        else gets JSON.
+
+        The document names this deployment's own origin in its `servers` block,
+        so it is built per origin rather than served as a fixed file.
+      security: []
+      responses:
+        "200":
+          description: AsyncAPI description
+          content:
+            application/vnd.aai.asyncapi+json:
+              schema:
+                type: object
+            application/vnd.aai.asyncapi+yaml:
+              schema:
+                type: string
+
+  /api/asyncapi.yaml:
+    get:
+      operationId: getAsyncApiDocYaml
+      tags: [Documentation]
+      summary: AsyncAPI description as YAML
+      description: |
+        The same document as `/api/asyncapi`, always as YAML, keeping the
+        source's comments and key order.
+      security: []
+      responses:
+        "200":
+          description: AsyncAPI description
+          content:
+            application/vnd.aai.asyncapi+yaml:
               schema:
                 type: string
 
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -6758,8 +6758,14 @@ components:
           type: string
         current:
           type: number
+          description: >-
+            The measured value. Absent when nothing was measured; present and
+            zero when the measurement was zero.
         configured:
           type: number
+          description: >-
+            The configured value. Absent when the finding is that nothing is
+            configured.
         suggested:
           type: number
         fixAction:
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -14,7 +14,7 @@ info:
     Error responses follow RFC 9457 (Problem Details).
 
     Every response (except `/-/` meta endpoints) includes:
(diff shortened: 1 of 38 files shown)


docs/configuration.mdx
```diff
@@ -81,7 +81,7 @@ the config file.
...
+  Comma-separated IP addresses or [CIDR](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) ranges of reverse proxies whose forwarding headers are trusted, e.g. `10.0.0.0/8,172.16.0.0/12`. Both [`Forwarded`][rfc7239] and `X-Forwarded-For` are read for the real client IP, the standard header first. Required for [header-based authentication][trusted-proxy-headers], and for `cert` mode behind a [TLS-terminating proxy][cert-behind-proxy]. When unset, every request counts as a direct connection, and `X-Forwarded-Proto`/`X-Forwarded-Host` are ignored — so absolute URLs (feed links and identifiers, OpenSearch templates, the API catalogue) name the address Cetacean was reached on inside the cluster. Set [`server.public_url`][server.public_url] instead if you would rather pin those URLs outright.
```
Radiergummi: This isn't interesting to users. Revert your changes.

docs/api.md
```diff
@@ -687,10 +699,12 @@ Every response outside the `/-/` meta endpoints carries [RFC 8631](https://www.r
...
 
 ```http
-Link: </api>; rel="service-desc", </api/context.jsonld>; rel="describedby", </.well-known/api-catalog>; rel="api-catalog"
+Link: </api>; rel="service-desc"; type="application/json", </api/asyncapi>; rel="service-desc"; type="application/vnd.aai.asyncapi+json;version=3.0.0", </api/context.jsonld>; rel="describedby", </.well-known/api-catalog>; rel="api-catalog"
 ```
 
-`service-desc` points at the OpenAPI spec, `describedby` at the JSON-LD context document, and `api-catalog` at the
+There are two `service-desc` links because there are two descriptions: the OpenAPI document at `/api` describes the
+request/response API, and the AsyncAPI document at `/api/asyncapi` describes the event streams. The `type` parameter
+tells them apart. `describedby` points at the JSON-LD context document, and `api-catalog` at the
```
Radiergummi: Why do we need to state this here?

docs/api.md
```diff
@@ -539,6 +539,18 @@ curl -H "Accept: text/event-stream" "http://localhost:9000/metrics?query=up&step
...
+`GET /api/asyncapi` describes every stream above as an [AsyncAPI 3.0](https://www.asyncapi.com/) document —
+`/api/asyncapi.json` serves the same thing. It names all twenty channels, the messages each can carry, and which of
+three cursor dialects its `id:` belongs to: a monotonic history id on the resource streams, an RFC 3339 timestamp that
+only moves forward on the log tails, and no id at all on `/metrics`. Paste it into
+[AsyncAPI Studio](https://studio.asyncapi.com/) to browse it; Cetacean ships no renderer of its own.
+
+Two things the document states that are easy to get wrong from the examples alone: a `batch` frame's payload is an
+**array** of envelopes rather than one, and a **replayed** event has no `resource` — replay reads the change history,
+which stores identity and not payload.
+
```
Radiergummi, line 550: This should be rephrased to be more to the point and terser

frontend/index.html
```diff
@@ -16,9 +16,15 @@
       rel="apple-touch-icon"
       href="./apple-touch-icon.png"
     />
+    <!--
+      use-credentials: a manifest is fetched with credentials omitted by
+      default, and /manifest.webmanifest is not exempt from authentication, so
+      without this the browser never gets one to install under auth.mode=oidc.
+    -->
```
Radiergummi: Who is this comment for? How is it helpful?

internal/recommendations/types.go
```diff
@@ -49,8 +49,10 @@ type Recommendation struct {
 	TargetName string   `json:"targetName"`
 	Resource   string   `json:"resource,omitempty"`
 	Message    string   `json:"message"`
-	Current    float64  `json:"current,omitempty"`
-	Configured float64  `json:"configured,omitempty"`
+	// Current and Configured are pointers for the same reason Suggested is:
+	// nil is "not measured", and a service using no CPU measures zero.
```
Radiergummi: That comment isn't helpful here

internal/api/csrf.go
```diff
@@ -56,3 +61,26 @@ func crossOriginProtection(cfg *CORSConfig, publicURL string) Constructor {
...
+// bearer token — and never from a credential the browser attaches on its own.
+//
+// There is no ambient authority for a cross-origin page to borrow, so the
+// protection defends nothing here and costs the thing the MCP authorization
+// profile depends on: a browser-based client completing a token exchange from
+// its own origin. RFC 7591 registration is the case that cannot be configured
+// around, since its whole point is a client the operator has never heard of.
+//
+// Deliberately not /oauth/authorize: consent runs under the user's session
+// cookie, which is exactly the ambient credential this protects.
```
Radiergummi: Prune this comment aggressively

internal/api/asyncapi.go
```diff
@@ -0,0 +1,94 @@
...
+// negotiate and no SPA fallback to reach.
+//
+// AsyncAPI 3.0 requires host on a server object, so the document names the
+// deployment's own origin — from the validated origin, never from a raw
+// forwarding header. The body is therefore a pure function of scheme, host and
+// base path, and one rendering is retained under that key so this document is
+// hashed and compressed once like every other served document. A single slot
+// rather than a map: origin falls back to r.Host when server.public_url is
+// unset, so a hostile client can vary the key without bound — a map would
+// grow, a slot degrades to rebuilding per request and no worse.
```
Radiergummi: Prune this comment aggressively

api/openapi.yaml
```diff
@@ -13,8 +13,15 @@ info:
     Detail responses include `@context`, `@id`, and `@type`.
     Error responses follow RFC 9457 (Problem Details).
 
+    Both descriptions are served as JSON and as YAML. `/api` and `/api/asyncapi` negotiate on `Accept`
+    (`application/yaml`, `text/yaml`, or the registered `application/vnd.oai.openapi` and
+    `application/vnd.aai.asyncapi+yaml`), and `/api/openapi.yaml` and `/api/asyncapi.yaml` address the
+    YAML form directly.
+
```
Radiergummi: Not helpful. Content negotiation just works, and the endpoints are listed below already. No need to restate it in prose.

internal/api/csrf.go
```diff
@@ -56,3 +61,17 @@ func crossOriginProtection(cfg *CORSConfig, publicURL string) Constructor {
 		})
 	}
 }
+
+// carriesItsOwnProof names the endpoints that authenticate from the request
+// body, with no ambient credential for a cross-origin page to borrow — so the
+// protection defends nothing and only breaks browser-based MCP clients.
+// /oauth/authorize is excluded: consent runs under the session cookie.
```
Radiergummi: This comment is too verbose and also confusing to read.

PR #236, opened 2026-09-12
feat(api): an entry point at the origin, and a catch-all that answers what clients accept

Radiergummi:
A path no route matches reaches the SPA fallback, which has exactly one representation. It answered every such request with the dashboard shell:

```
GET /nonesuch.json                       ->  200 text/html
GET /nonesuch   Accept: application/json ->  200 text/html
```

A client that named JSON was told its request had succeeded and handed a web page. Every registered well-known URI whose name ends in `.json` is unreachable for the same reason — `assetlinks.json`, `did.json`, `host-meta.json`, `gpc.json` and eleven others in the IANA registry — and `frontend/dist/assets-manifest.json` already is. `/-/sbom.cdx.json` had to be registered as `/-/sbom.cdx` to work around it.

## The root cause

The shell is `text/html`, so the question the fallback has to ask is **whether the client will accept `text/html`**. It could not ask it.

`negotiate` collapses three distinct client states — an extension suffix, an explicit `Accept`, and no preference at all (`*/*` or an absent header) — into a single `ContentType`. `Accept: */*` and `Accept: application/json` both arrive as `ContentTypeJSON`, and only one of them rules out the shell. The information needed to decide was destroyed one layer up.

`negotiate` now answers both questions from the one parse it already performs. `rangesAcceptHTML` implements RFC 9110 §12.5.1: no preference admits everything, and the most specific range matching a type carries the weight that applies. `parseAccept` splits into `parseAcceptRanges` and `bestMatch` to expose the ranges. The verdict is recorded only when HTML is *refused*, so the zero value is permissive and the allocation lands only on requests that were going to 404 anyway.

## Two things the fix uncovered

**`bestMatch` ranked by weight but never excluded on it.** §12.5.1 defines `q=0` as *not acceptable*, so `Accept: text/html;q=0` resolved to `ContentTypeHTML` while `rangesAcceptHTML` correctly refused it — one parse yielding two answers that disagreed about the same client. `bestMatch` now skips a zero-weighted range, and such a request is answered `406`.

**The feed-probe defence was in the wrong handler.** `hasMidPathExtension` refuses `/nodes.atom/feed`, a feed reader probing for a feed. It lived in `NewSPAHandler` — but `spa` is not only the catch-all, it is the `text/html` arm of **forty** `contentNegotiated` endpoints, so a "this is never a client-side route" path heuristic ran where that premise is false by construction. Narrowing it from *any dot* to the `extensionTypes` table was not enough: a service named `web.json` or `web.atom` then 404'd its own detail tabs — the very bug the narrowing was meant to fix, since Docker permits dots in names and `/services/:id/:subResource` is a real route.

It now sits in the catch-all, the one route that knows nothing matched. Its plain-text `http.NotFound` becomes the RFC 9457 problem document the rest of the API uses.

## Behaviour

```
GET /nonesuch.json           Accept: text/html                    ->  404   suffix overrides Accept
GET /nonesuch                Accept: application/json             ->  404   newly possible
GET /nonesuch                Accept: */*, text/html;q=0           ->  406   refused by weight
GET /nodes.atom/feed         Accept: */*                          ->  404   discovery probe
GET /services                Accept: text/html;q=0                ->  406   was 200 text/html
GET /nonesuch                Accept: application/json, */*;q=0.1  ->  SHELL a preference is not an exclusion
GET /nonesuch                Accept: */*  |  (no Accept)  |  text/*  ->  SHELL
GET /nonesuch.html           Accept: text/html                    ->  SHELL
GET /services/web.api/logs   Accept: text/html                    ->  SHELL was 404
GET /services/web.json/logs  Accept: text/html                    ->  SHELL was 404
GET /assets/app-abc.js                                            ->  200   unchanged
GET /favicon.ico             Accept: image/png                    ->  200   the file, was 406
GET /nonesuch                Accept: image/png                    ->  404   was 406
GET /                        Accept: application/json             ->  200   the entry point, was the shell
GET /                        Accept: text/html                    ->  200   the dashboard, unchanged
GET /index.json                                                   ->  301   Location: /.json
```

Tests cover these, plus a focused table for `rangesAcceptHTML` — the RFC rule is the part worth pinning, since the resolved `ContentType` cannot express it.

## The 406 that fired before the lookup

The catch-all also refused `ContentTypeUnsupported` with `406` *before* anything looked at the disk, so `GET /favicon.ico` with `Accept: image/png` answered *this endpoint supports text/html* about a file sitting right there.

That check asked the API's media-type table to recognise an `Accept` on the one route that serves **static files**, whose types the table does not enumerate and never will. `image/png` is not a type nothing serves; it is a type `negotiate` has no opinion about. It was inconsistent on its own terms, too — `Accept: text/vnd.graphviz` resolves to `ContentTypeDOT`, passed the gate, and got handed JavaScript.

Nothing replaces it. The check that matters is the one this PR already put below the `fs.Stat`, where a file can be told from a client-side route, and `rangesAcceptHTML` refuses `image/png` there. So the route answers the file or `404`, and never `406`: what it has is one representation with nothing to choose between, and what it has not is no resource to hold representations at all — exactly the case RFC 9110 §12.1 lets a server disregard `Accept` for. Honouring it instead would cost the parsed ranges in the context and a decision inside `serveAsset`, to tell a client that asked for a favicon that it cannot read images.

`TestUnservedTypeIsRefusedByTheEndpoint` loses its `/not-a-route` probe: it pinned the removed behaviour, and the endpoints it still covers are the ones that genuinely choose among representations.

## What the origin is

Fixing the two above raised the question underneath them, and it turned out the repo already disagreed with itself about the answer.

`internal/api/apicatalog.go` has anchored the web API at `/` since it was written:

```go
restAPI := link("/")
items := []linkset.Target{{Href: restAPI, Title: "Cetacean web API"}}
```

with `service-desc`, `service-doc`, `describedby` and `status` hung off that anchor. So the RFC 9727 catalog tells every client *the API is at `/`* — and `/` handed it a web page. Meanwhile `internal/contract/excused.go` recorded the opposite in as many words: `"/": "serves the embedded SPA and /assets/*; not an API surface"`.

`GET /{$}` makes the catalog right. `text/html` still reaches the dashboard, which is what a browser at the origin wants; JSON gets a JSON-LD document naming every top-level collection, so a client that knows only where a deployment lives can find the rest from there.

```json
{
  "@context": "/api/context.jsonld",
  "@id": "/",
  "@type": "EntryPoint",
  "name": "Cetacean",
  "resources": {
    "cluster": {"@id": "/cluster"},
    "nodes": {"@id": "/nodes"},
    "services": {"@id": "/services"},
    "…": "…"
  },
  "version": "0.11.0"
}
```

`{$}` anchors the pattern to the exact path, so the catch-all below still answers everything else. Each link is a node reference rather than a bare string — under `@vocab` a string is a literal, not a link.

**Not a redirect to `/api`.** That is the interface description, not a way into the data: it answers "find me the services" with 400KB of schema, and it is already advertised as `service-desc` on every response. Redirecting to the catalog would be circular — this API is one `item` in it.

**`/index` redirects, keeping the suffix.** `negotiate` strips the suffix before routing, so one route covers `/index`, `/index.html` and `/index.json`; the suffix goes back on the target, which would otherwise be re-negotiated from an `Accept` that may disagree. That is what the stripped suffix is for, so `negotiate` now records it.

**Exempt from the readiness gate, not from authentication.** The document is static — a client probing a server whose Docker is unreachable is exactly when the map is worth most. The catalog is unauthenticated and points here; this is not. Discovery is public, the API behind it is not.

`TestEveryOperationIsGatedAtItsDeclaredTier` compares router patterns against spec paths, which spell the root differently. `{$}` is ServeMux syntax rather than a path parameter, so the comparison drops it.

## Terminology

"REST API" is now "web API" throughout. What this serves is content-negotiated HTTP with JSON-LD identifiers and link relations — more hypermedia than most things called REST, and still not the thing Fielding described.

Only that phrase changes. "REST" alone stays where it names this transport against MCP, which `internal/mcp` and `internal/cluster` do constantly and exactly. The archived designs under `docs/specs` keep their wording: they record what was decided at the time. One occurrence is a served response body — the linkset's title for the member API — so the spec's example moves with it.

## Context

Found reviewing #233, whose `jwks_uri` sits at `/oauth/jwks` partly to sidestep this. That path stays correct on its own merits — `jwks.json` is not an IANA-registered well-known URI suffix (neither `jwks` nor `jwks.json` appears in the registry), RFC 8615 §3 requires registration, and RFC 8414 leaves `jwks_uri` a free-form URL discovered through the metadata document. It simply should not have needed the workaround. Worth dropping the negotiation clause from the comment there once this lands.

## Verification

`main` is merged in, resolving the conflict (`CHANGELOG.md` only). `make fmt-check` now passes too, website files included. `go test ./...` passes, `internal/api` also under `-race`; `golangci-lint run ./...` reports 0 issues; `gofmt -l .` is empty.

https://claude.ai/code/session_014RyDznPwpSWDQWBojnT4i6


internal/api/negotiate.go
```diff
@@ -128,6 +149,22 @@ var extensionTypes = []struct {
 	{".csv", ContentTypeCSV},
 }
 
+// hasMidPathExtension reports whether a known extension suffix appears in a
+// non-terminal segment, which a segment ends with exactly when the path holds
+// it followed by a separator. resolveExtension answers this for the terminal
+// segment. Matching the table rather than any dot is what keeps a Docker name
+// like web.json from reading as a representation; the router calls this only
+// where no route matched, since where one did, it is a real resource.
```
Radiergummi: Prune this aggressively.

internal/api/negotiate.go
```diff
@@ -142,6 +179,32 @@ func resolveExtension(r *http.Request) ContentType {
 	return ContentTypeUnsupported
 }
 
+// rangesAcceptHTML reports whether the ranges admit text/html at a usable
+// quality, per RFC 9110 §12.5.1: no preference admits everything, and the most
+// specific range matching a type is the one whose weight applies — so
+// "*/*, text/html;q=0" refuses HTML while "application/json, */*;q=0.1" does
+// not. Answering this from the resolved ContentType is impossible: it reports
+// JSON for a wildcard and for an explicit application/json alike.
```
Radiergummi: Prune this aggressively.

internal/api/router.go
```diff
@@ -823,6 +823,16 @@ func newRouter(cfg RouterConfig) (http.Handler, []string) {
 			return
 		}
 
+		// A known suffix in an inner segment (/nodes.atom/feed) is a feed
+		// reader's discovery probe, not a client-side route. It is decided
+		// here because only this route knows nothing matched: `spa` is also
+		// the text/html arm of every contentNegotiated endpoint, where a
+		// dotted Docker name like web.json is a real, routed resource.
```
Radiergummi: Prune this aggressively.

internal/api/spa_test.go
```diff
@@ -415,3 +415,121 @@ func TestEmbeddedManifestOmitsIndexHTML(t *testing.T) {
 		}
 	}
 }
+
+// The catch-all serves one representation, text/html. A client that ruled it
+// out cannot be answered with it, however it said so: negotiate strips an
+// extension suffix and mutates r.URL.Path before routing, so a path no route
+// matches arrives with the suffix gone and only the negotiation record left.
+// Registered well-known URIs ending in .json (assetlinks.json, did.json,
+// host-meta.json) all land here.
```
Radiergummi: Prune this aggressively.

PR #237, opened 2026-09-12
test: drive Cetacean end to end, and fix what that exposes

Radiergummi:
Builds a black-box end-to-end harness that runs the real `./cetacean` binary against a real
Docker-in-Docker swarm, and fixes the defects that driving it exposed.

The harness is the means, not the point. **29 product fixes** are in here, and essentially none of
them were reachable from a unit test: they are wiring, protocol conformance, lifecycle and failure
behaviour — the things that only go wrong when an actual process talks to an actual engine over an
actual socket.

## What it found

A representative sample, not the full list:

| Defect | Why it mattered |
|---|---|
| Distributed tracing never exported a span | The documented endpoint (`http://collector:4318`) posted to `/` instead of `/v1/traces`. Startup logged "tracing enabled" and nothing ever arrived — the feature had never worked as documented |
| Graceful shutdown abandoned the drain | `main` returned the moment the listener closed while `Shutdown` was still draining. Every rolling update truncated in-flight requests, including writes already at the Docker API |
| A lost Docker connection was invisible | Frozen cache served as "Live", readiness still passing, no metric. For an observability tool, the worst way to fail |
| Docker Engine < 27 fails every request | Symptoms the troubleshooting guide misattributes to an unmounted socket. The requirement was documented nowhere |
| `If-Match` evaluated against the cache | A validator compared against an asynchronously-filled cache can only refuse a change the cache already saw |
| Service patches merged against the cache | Merging into the cache drops whatever was written in between, and nothing downstream could catch it |
| A write to an unrouted path answered 200 + HTML | A client checking only the status read a failed write as a success |
| `POST /-/resync` was unauthenticated | Anyone who could reach the port could amplify one request into a full cluster enumeration |
| Refresh-token theft detection never ran | The replay was refused as unknown before detection could fire — on every refresh in the default configuration |
| Identifiers ignored `base_path` | Every `@id` a listing handed out 404'd, beside a correctly prefixed `@context` in the same document |

## What it proves

Three properties now have evidence rather than assertion:

- **No data races** in the real server under real Docker events. `make test-stack-race` drives every
  lane against a `-race` binary with `halt_on_error=1`. CI's `go test -race ./...` covers the
  packages in isolation; nothing covered the watcher goroutines, the cache mutex, the SSE fan-out,
  the MCP notification manager and the ACL hot reload working together.
- **No goroutine leak**: 17 before, 16 after 120 streaming connections came and went.
- **SSE fan-out isolation**: 120 of 120 events delivered to a healthy subscriber past four stalled
  clients.

## Shape

- `test/e2e/` — 25 lanes, each owning a reserved port, driving HTTP, SSE and MCP. Nine *gate* tests
  fail if the app grows a route, tool, representation, filter field or notification type that
  nothing drives, so the suite cannot silently fall behind the product.
- `frontend/e2e/` — the existing Playwright suite, now reproducible against the harness environment.
  Ten specs that could never run are fixed; the rule is **ask the server, not the DOM**.
- Never runs in CI: it needs a privileged container and takes minutes. `make test-stack` locally.

## Reviewing this

175 files and ~32k lines, so **read it commit by commit** — each product fix is one commit carrying
its own test and its own changelog entry, and every commit builds and vets independently.

Two judgement calls are worth disagreeing with if you want to:

- **Readiness deliberately keeps passing while the Docker engine is down.** The container
  healthcheck polls `/-/ready`, so reporting a dead engine as unready would restart a process that
  is working correctly and would come back identical. Staleness is reported on `/-/health` and
  `/-/metrics` instead.
- **The Docker API stays pinned, not negotiated.** Negotiating down would leave Cetacean partially
  working against an older daemon, answering with whatever that API returns and saying nothing about
  what it could not ask for.

## Verification

`make check`, `make test-stack` (all lanes, 66% coverage), `make test-stack-race` (0 races, 0 unclean
exits) and the browser suite (159 passed, 3 skipped, repeatably across seven runs) all green.

One browser spec is recorded in the README as a known flake — it failed once in eight full runs and
could not be reproduced. It is documented rather than papered over with a retry, because masking it
would repeat exactly the mistake that left ten specs silently dead.

https://claude.ai/code/session_01QHa88tnVJVw5B62kXntMPU


docs/api.md
```diff
@@ -297,6 +225,31 @@ wrap items as `{ items, total, limit, offset }` with [RFC 8288](https://www.rfc-
...
+
+Volumes and stacks are keyed by name already, so their paths never redirect. Tasks are addressable by ID only: a
+task's `<service>.<slot>` name is derived from its parent rather than stored on it.
+
+A name matching more than one resource is answered `409 Conflict` with [`API014`](api/errors#API014), whose detail
+lists every ID the request could have meant. Swarm enforces unique names for services, configs, secrets and networks
+but not for node hostnames, so nodes are the type where this arises.
+
+Requests negotiating `text/html` are never redirected — those paths are the [dashboard][dashboard]'s own routing
+surface.
```
Radiergummi: Prune this section. It contains too much low-level detail irrelevant to users.

docs/api.md
```diff
@@ -575,7 +528,17 @@ There is no general rate limiting. Concurrent streams are capped, and a request
...
+`/-/` endpoints it requires authentication and a grant, because each call sweeps the whole Docker API,
+but it is not gated on the operations level — it re-reads the cluster and never changes it.
+
+`GET /-/health` carries a `watcher` object alongside the version fields, reporting whether Cetacean is
+still tracking the cluster — `connected` for the Docker event stream, and `lastSyncAt` /
+`lastSyncAgeSeconds` for the last successful read. It matters because losing that connection is
+invisible from every other endpoint: they keep answering from the cache, and `/-/ready` keeps
+returning `200` so an orchestrator does not restart a process that would come back identical. The same
+facts are on `/-/metrics` as `cetacean_watcher_connected`, `cetacean_cache_last_sync_timestamp_seconds`
+and `cetacean_cache_sync_failures_total`.
```
Radiergummi: Again, how is most of this relevant to users here?

docs/api.md
```diff
@@ -649,14 +614,21 @@ A `412` always means the resource moved. Where the current representation cannot
 `DELETE /plugins/{name}` inspects the daemon rather than the cache — the write answers `503`
 (`ENG001`) or `500` (`ENG004`) instead, so an unreachable daemon is not reported as a stale `ETag`.
 
-30 endpoints support it: `PATCH /services/{id}/env`, `PATCH /services/{id}/labels`,
+The comparison is made against the Docker engine, not against the cached copy a `GET` is served
+from: the cache is filled asynchronously, so a validator compared against it could only ever
+refuse a change the cache had already seen — never the lost update the header exists to prevent.
+Evaluating the condition therefore also brings the cached copy up to date, which is what lets a
+`412` be recovered from in a single round trip: re-read the resource and the `GET` answers the
+validator the write was held to.
```
Radiergummi: And one more time, also too much irrelevant detail.

docs/authorization.md
```diff
@@ -57,6 +58,11 @@ Two inheritance rules widen a grant beyond a literal match:
 Task patterns match task IDs, which change every time a replica is replaced. Grant the parent service or the stack
 instead of naming tasks.
 
+A backslash escapes the character after it, so `service:release-\*` grants exactly the service named
+`release-*` rather than every service whose name begins `release-`. It is significant whether you meant it
+that way or not: `service:a\b` grants `ab`, not `a\b`. Docker resource names cannot contain a backslash,
+so a pattern carrying one is always either an escape or a mistake.
+
```
Radiergummi: Remove this.

docs/getting-started.md
```diff
@@ -10,6 +10,9 @@ tags: [ installation, docker, swarm, quickstart ]
 ## Requirements
 
 - A Docker Swarm Mode cluster. Single-node swarms work.
+- **Docker Engine 27.0 or newer.** Cetacean speaks Docker API 1.46, and an older daemon refuses every
+  request as too new. It says so at startup and stops, naming both versions, rather than coming up and
+  serving empty pages.
```
Radiergummi:
```suggestion
- **Docker Engine 27.0 or newer, speaking Docker API 1.46 or above.**
```

docs/getting-started.md
```diff
@@ -122,6 +125,25 @@ change at all. Set it to `0` for a read-only deployment.
...
+```json
+{ "status": "ok", "watcher": { "connected": false, "lastSyncAgeSeconds": 942.1 } }
+```
+
+`connected` is whether the event stream is established; `lastSyncAgeSeconds` is how long ago
+Cetacean last read the cluster successfully, against a re-sync that runs every five minutes. The same
+facts are on [`/-/metrics`][api.selfmetrics] as `cetacean_watcher_connected`,
+`cetacean_cache_last_sync_timestamp_seconds` and `cetacean_cache_sync_failures_total` — alert on the
+first being `0`, or on the second falling far enough behind `time()`.
+
```
Radiergummi: Irrelevant. If there is a programmatic way to discover the engine has gone, the dashboard should check this and tell the user about it - they should not have to consult the docs to learn that might be the reason for a stale dashboard and then query a JSON endpoint to prove it.

docs/mcp.md
```diff
@@ -106,6 +106,8 @@ fails—so you can see which agent scaled what, and when. If the calling agent i
 picked up and Cetacean's spans join the same trace, putting the agent's turn and the Docker call it produced on
 one timeline.
 
+Give it the collector's base URL; Cetacean appends OTLP's `/v1/traces` itself, so a collector behind a gateway prefix (`https://gateway.example.com/otlp`) works as written. An endpoint that already ends in `/v1/traces` is left as it is.
```
Radiergummi: If it doesn't matter whether the user appends the path, why bring this up at all?

Radiergummi:
Superseded by a three-PR stack, split out of this branch:

- #239 — `test:` the e2e harness and the contract inventories (86 files, +26.2k). Compiles and runs against unmodified `main`; three invariants are red there and are recorded as defects rather than weakened.
- #240 — `fix:` the defects the harness exposed (+8.4k/−0.9k). Deletes all three excuse entries #239 registers.
- #241 — `style:` the comment-budget pass (267 files, +2784/−6363), mechanically verified to contain zero non-comment lines.

The split is lossless: the tip of #241 is byte-identical to this branch's head apart from four deliberate `test/e2e` lint fixes (`bodyclose` ×2, `modernize` ×2) that close the `golangci-lint run --build-tags=e2e` gate this branch was failing.

Closing in favour of the stack. The granular history — 123 commits, one invariant per commit per the campaign handover — stays on `test/coverage-campaign`, which is not deleted.

PR #244, opened 2026-09-13
refactor(oauth)!: extract the authorization server, and make it opt-in

Radiergummi:
Two design investigations, then the first piece of work they call for.

## Breaking changes

| was | now |
|---|---|
| implicit (MCP + any auth mode but `none`) | `oauth.enabled` / `CETACEAN_OAUTH_ENABLED`, default **false** |
| `mcp.issuer`, `mcp.signing_key`, `mcp.access_token_ttl`, `mcp.refresh_token_ttl`, `mcp.consent_ttl` | `oauth.*` / `CETACEAN_OAUTH_*` |
| `[mcp.oauth]` (`require_resource_indicator`, `dcr_*`, `cimd_enabled`) | `[oauth]` |
| `mcp.oauth.auth_bypass` | `mcp.auth_bypass` (stays MCP's) |
| `{data_dir}/mcp-tokens.json` | `oauth-tokens.json`, old name read once on first start |

Access tokens issued before this stop verifying and the published `kid` changes, because the key-derivation labels moved too. Refresh tokens survive, so clients refresh once rather than re-authorizing — the upgrade shape the unreleased ES256 entry already describes, now folded into one instruction.

## Why

`docs/specs/2026-09-13-api-oauth-tokens-design.md` plans to make the server issue tokens for the REST API as a second protected resource, which needs it shared rather than MCP's. Extracted first, the diff that adds a second audience is reviewable; done together it would be buried in a rename.

It also becomes opt-in, which was the repo owner's call and turned out to simplify two things rather than complicate them. `OAuthIssuerRequired` used to approximate "will anything drive the authorize flow" from the auth mode and the bypass list; with an explicit flag it *is* the flag, and `internal/config`'s OAuth half now names MCP nowhere. And the opt-in decouples the server's lifecycle from `mcp.enabled` by itself, which the design doc had deferred to a later change.

## The part worth reviewing carefully

**Three combinations are settled at startup** rather than discovered later. MCP under any auth mode but `none` *without* the server is refused, because `/mcp` is exempt from the auth middleware and authenticates itself — with no server there is no bearer check, and the endpoint would serve the cluster's write surface unauthenticated. `mcp.auth_bypass` does not excuse it, and the error says so: the bypass runs *inside* the very middleware an absent server never installs. The server under `none` is refused too (consent would ask an anonymous user to approve a client). The server with nothing consuming it is allowed and warned about.

**The nil-interface guard in `main.go`.** `Options.OAuth` became an interface, and `oauthSrv` is nil when no server was built. A nil pointer inside an interface is not nil, so the existing check in `Handler()` would have passed, `bearerAuth` would have armed, and every unauthenticated `/mcp` request would have panicked on a nil receiver. `govet`'s nilness does not model typed-nil-in-interface and `main.go` has no tests, so the guard follows the precedent the Prometheus client already set six lines above.

**Both golden constants were recomputed from the curve parameters**, not copied from the failing test output — `TestGoldenKID` exists to catch drift in go-jose's JWK encoding, and pasting whatever the code now produces would make it a tautology. The method was validated by reproducing the two committed values first.

## Shape of the series

Eight commits, each green. `git mv` is content-free so all 27 files report `R100` and `git log --follow` still works across the boundary; the renames land after, on files git has already learned the new path for. The first commit only pins behaviour the later ones change — the identity a bearer token yields, the `WWW-Authenticate` value, the CSRF cookie's wire name — so each effect shows up as a one-line diff rather than an absence.

`internal/mcp`'s *production* code no longer imports the package; its tests still do, deliberately, because they mint real ES256 tokens and are the only proof verification works end to end.

`internal/oauth/independence_test.go` now enforces the point of the exercise by scanning the package's own source for its consumer's name. It carries one documented exception, which it found itself: the migration path has to name the file it migrates from.

## Verification

`make test` and `make typecheck` are green (the local `golangci-lint` is built against Go 1.25 and refuses a 1.26.6 target, so CI owns `lint`/`fmt-check`). Live checks: both discovery documents serve with the right issuer and resource, the state-file migration fires and logs, and all three validation refusals behave. `/oauth/jwks` and `/mcp` could not be exercised here — they sit behind a Docker-reachability gate and this container has no daemon — so the `kid` change and the unguarded-`/mcp` path rest on `TestGoldenKID` and `TestHandlerWithoutOAuthServesUnguarded`.

## Also on this branch

The two investigations that prompted it: `2026-09-13-native-apple-client-design.md` (a native macOS/iOS client — feasible, with credentials as the blocking gap and multi-cluster as the case the web dashboard structurally cannot make) and `2026-09-13-api-oauth-tokens-design.md` (Phase 0 in detail: resource identifiers, the RFC 9728 metadata layout, and where token verification runs). Decisions 6 and 7 of the latter are annotated where this work went the other way.

https://claude.ai/code/session_01BByYVLBwsLhty9aFdoSPUy

### DIFF
--- a/docs/specs/2026-09-13-api-oauth-tokens-design.md
+++ b/docs/specs/2026-09-13-api-oauth-tokens-design.md
@@ -1,0 +1,279 @@
+# API Access Tokens from the OAuth Server
+
+**Date:** 2026-09-13
+**Status:** Investigation. Phase 0 of `2026-09-13-native-apple-client-design.md`, written up
+separately because it stands on its own: any CLI, script or third-party client wants it whether or
+not a native app is ever built.
+
+## Summary
+
+Teach the existing OAuth 2.1 authorization server to issue access tokens for the REST API as a
+second protected resource, and teach the auth middleware to accept them. The API gains a credential
+a client can obtain unattended, on every auth provider, with no operator configuration.
+
+The two resources stay separate: a token minted for MCP must not open the REST write surface, and
+vice versa. That separation is the whole reason to do this through resource indicators rather than
+by widening the existing audience.
+
+## What exists today
+
+`internal/mcp/oauth` is a complete AS: RFC 7591 DCR, CIMD, PKCE S256 (required, with verifier
+length and alphabet enforced), authorization codes with a 60 s TTL, rotating refresh tokens with
+theft detection, a consent screen with remembered approvals, RFC 8707 resource indicators, RFC 9068
+`at+jwt` access tokens signed ES256 with a published JWK Set, RFC 9728 protected resource metadata,
+and durable state under `{data_dir}/mcp-tokens.json`. Defaults: access 1 h, refresh 30 d, consent
+90 d.
+
+Five facts about it shape everything below.
+
+- **The audience is a constant.** `TokenIssuer.Audience` is set once from `ServerConfig.MCPResource`
+  and `IssueAccessToken` stamps it on every token; `VerifyAccessToken` compares for exact equality.
+  The resource indicator *is* validated and *is* stored on the authorization code and the refresh
+  token, but nothing reflects it into `aud` — with one resource there was nothing to reflect.
+- **There are no scopes.** `AccessTokenClaims` is `sub`, `groups`, `client_id`. Authorization comes
+  entirely from the upstream identity reaching the ACL. A token therefore carries exactly its
+  user's access, no more and no less.
+- **The AS exists only when MCP does.** `setupMCP` returns early on `!cfg.MCP.Enabled`, and
+  constructs the server only when `authMode != "none"`. Its knobs live in `config.MCPConfig`.
+- **`internal/api` does not import `internal/mcp`.** Deliberate, and stated at `carriesItsOwnProof`
+  in `internal/api/csrf.go`. The MCP handler and the OAuth routes both reach the router as
+  injected values from `main.go`.
+- **The resource server is `internal/mcp`, not the AS.** `Server.bearerAuth` verifies the token and
+  builds an `auth.Identity` from the claims. Nothing else in the tree verifies a Cetacean-issued
+  token.
+
+## Decisions
+
+### 1. Resource identifiers, and where the PRM documents live
+
+This is the part that actually needs thought, because RFC 9728 derives the metadata URL from the
+resource identifier's path, and today's single document sits at the root-path location while
+describing a resource that has a path.
+
+Today:
+
+| | |
+|---|---|
+| Resource identifier | `{issuer}{base_path}/mcp` |
+| PRM served at | `{issuer}{base_path}/.well-known/oauth-protected-resource` |
+| `WWW-Authenticate` points at | that same URL |
+
+RFC 9728 §3.1 builds the metadata URL by inserting `/.well-known/oauth-protected-resource` after
+the host and appending the resource's path — so the MCP resource's metadata belongs at
+`/.well-known/oauth-protected-resource/mcp`, and the root-path location belongs to a resource
+identifier with no path.
+
+**Recommendation.** The API's resource identifier is the deployment root, `{issuer}{base_path}`,
+and takes the root-path PRM. MCP keeps its identifier and moves to the path-based one:
+
+| Resource | Identifier | PRM |
+|---|---|---|
+| REST API | `{issuer}{base_path}` | `{base_path}/.well-known/oauth-protected-resource` |
+| MCP | `{issuer}{base_path}/mcp` | `{base_path}/.well-known/oauth-protected-resource/mcp` |
+
+Each `WWW-Authenticate` names its own document, so a client that discovers the way RFC 9728 and the
+MCP authorization spec tell it to — read the 401, follow `resource_metadata` — is unaffected by the
+move. A client that hardcoded the root path breaks. That is wire-visible and owes a changelog entry
+and a line in `docs/mcp.md`; it cannot be softened by serving both, because one URL can describe
+only one resource.
+
+**The root identifier does not imply containment.** A reader will assume a token for
+`https://host/` covers `https://host/mcp` because one path is under the other. It does not, and must
+not: audience comparison is exact string equality, so the two are simply different audiences. Say so
+in the doc comment, because the alternative reading is the audience-confusion attack the resource
+indicator exists to prevent — a user who grants an AI agent MCP access has not granted it
+`DELETE /services/{id}`.
+
+Rejected: giving the API the identifier `{issuer}{base_path}/api`. It collides with the OpenAPI
+endpoint, which the current PRM already names as `resource_documentation`, and it wrongly suggests
+the token is scoped to that path.
+
+### 2. `aud` comes from the bound resource
+
+`IssueAccessToken` takes the audience per call, from `codeData.Resource` / `RefreshTokenData.Resource`
+— the value already validated and stored. `VerifyAccessToken` takes the expected audience from its
+caller. `TokenIssuer.Audience` stops being a field, or becomes a default for the no-indicator case.
+
+`ValidateResourceIndicator` then needs a set rather than one expected value: empty and not required
+resolves to a configured default (the API resource, which is what an unmodified client should get),
+empty and required is an error as now, non-empty must match one of the known identifiers exactly.
+Keep the exactness — a prefix match here would reintroduce containment.
+
+This is the change that makes a mis-targeted token fail loudly: presenting an MCP-audience token to
+the REST API produces `ErrAudienceMismatch`, which the middleware must answer as a 401
+`invalid_token` naming the *API's* PRM, not as a fall-through.
+
+### 3. Verification runs in `internal/auth`, behind an interface
+
+`internal/api` cannot import `internal/mcp`, and putting the verifier in `internal/auth` while the
+OAuth package imports `internal/auth` for `IdentityFromContext` would be a cycle.
+
+Use the pattern the codebase already uses twice (`auth.SetErrorWriter`, `api.RouterConfig.MCPHandler`):
+`internal/auth` declares the narrow interface it needs, `main.go` wires the implementation.
+
+```go
+// In internal/auth.
+type TokenVerifier interface {
+    // VerifyAPIToken returns the identity a Cetacean-issued API token carries.
+    VerifyAPIToken(token string) (*Identity, error)
+}
+```
+
+`auth.Middleware` takes an optional verifier and consults it before the provider:
+
+1. No `Authorization: Bearer` → provider path, unchanged.
+2. Bearer present, verifier absent → provider path, unchanged.
+3. Bearer present and the token verifies → identity from claims, proceed.
+4. Bearer present, token is ours and does not verify → **401, final.** Never fall through, and
+   never the OIDC provider's `text/html` redirect-to-login branch: a native client must receive a
+   401 with `WWW-Authenticate`, not a 302.
+5. Bearer present and the token is not ours → provider path, so an OIDC ID token keeps working.
+
+**4 and 5 are the fiddly bit**, and the only real trap in this design. Two bearer dialects share one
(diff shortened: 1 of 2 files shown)


Radiergummi:
## CodeQL `go/path-injection` at `main.go:773` — relocated, not introduced

CodeQL flagged `os.MkdirAll(d.cfg.DataDir, 0700)` inside the new `setupOAuth`. I checked it rather than suppressing it, and it is the pre-existing data-dir creation, moved.

The diff removes one `MkdirAll` (from `setupMCP`) and adds one (to `setupOAuth`) — net zero sinks. Its twin is still on `main`, byte-identical, at `main.go:255-256`:

```go
snapshotPath = filepath.Join(cfg.DataDir, "snapshot.json")
//nolint:gosec // DataDir is operator-configured, not user input
if err := os.MkdirAll(cfg.DataDir, 0700); err != nil {
```

CodeQL attributes it to this PR because the lines are new *in the diff*, and it does not read `//nolint`.

On the substance: `DataDir` is `resolve(flags.DataDir, "CETACEAN_DATA_DIR", fDataDir, "./data")` (`internal/config/config.go:168`) — a CLI flag, an environment variable, or a TOML key. CodeQL's taint model counts those as user-controlled, which is true in the sense that *someone* supplies them; but that someone is the operator running the binary, who can already write anywhere the process can. There is no request-derived input on this path.

**I have not changed the code for it.** `filepath.Join` already cleans its result, so the only candidate edit is routing `DataDir` through `filepath.Clean` before `MkdirAll` — functionally a no-op, uncertain to satisfy the query, and it would sit awkwardly next to the comment recording the opposite judgment. Contorting the code to quiet a scanner that is right about the dataflow and wrong about the threat seemed worse than saying so.

Two ways to close it, both yours: dismiss the alert as "used in tests / not applicable" in the security tab, or add a `go/path-injection` filter to the CodeQL config if the twin at `main.go:255` is also open there. Happy to do the latter if you tell me which you prefer — I did not want to edit scanning configuration unasked.

---

Radiergummi:
## Correction: there were two CodeQL alerts, and one of them was mine

My comment above was written against alert 47 alone and said the finding was "relocated, not introduced". That holds for 47 (`main.go` / `os.MkdirAll`), but alert **48** landed at `internal/oauth/persist.go:245` and that one was new code I wrote. I should have waited for the Go analysis to finish before characterising it.

Fixed in `abadbee`. The legacy-path check was:

```go
func fileExists(path string) bool {
	_, err := os.Stat(path)
	return !errors.Is(err, fs.ErrNotExist)
}
```

— a genuinely new filesystem sink. The fallback now reads the current path first and keys off `readState`'s own `fs.ErrNotExist`, so `fileExists` is gone entirely. That removes the sink, removes a `stat`, and removes the window between checking for the file and reading it.

It also sharpened the rule rather than just relocating it. The old version treated an *unreadable* file as present via a negated `errors.Is`, which was easy to misread; now only a missing file falls back, and a corrupt one warns exactly as before. That branch had nothing holding it, so `TestCorruptStateDoesNotFallBackToTheLegacyPath` now does — falling back there would quietly restore state an operator had deliberately replaced.

**Alert 47 is unchanged and my assessment of it stands**: the `os.MkdirAll(cfg.DataDir, …)` in `setupOAuth` is the pre-existing data-dir creation relocated out of `setupMCP`, its twin is still on `main` at `main.go:255-256` with the same `//nolint:gosec`, and `DataDir` comes from a flag, `CETACEAN_DATA_DIR`, or a TOML key. Still yours to dismiss or filter; still happy to add the CodeQL config entry if you prefer that route.

Separately, `lint-go` went red on `78d4dd7` for a `golines` line-length issue my fixture rename caused, fixed in `84e528c`. The local `golangci-lint` refuses this repo's Go version, so I installed `golines` standalone — it now reports clean across every `.go` file, not just the changed ones.

---

Radiergummi:
## Follow-ups toward a native client

CI is green on `23d6f1b` — including CodeQL, which the data-dir consolidation cleared, so no dismissal is needed after all. Removing the duplicate `MkdirAll` was the real fix.

This PR is Phase 0's *first half*: the authorization server is standalone and opt-in, but it still serves exactly one resource. Here is everything left, in the order it wants doing. The two design docs on this branch carry the reasoning; this is the checklist.

### Phase 0, remainder — the API as a second protected resource

The change this extraction existed to make reviewable. From `docs/specs/2026-09-13-api-oauth-tokens-design.md`:

- [ ] **Resource identifiers and PRM layout.** API takes the deployment root `{issuer}{base_path}` and the root-path PRM; MCP keeps its identifier and moves to `/.well-known/oauth-protected-resource/mcp` per RFC 9728 §3.1. Wire-visible — conformant clients follow `resource_metadata` from the 401, but it owes a changelog entry and a line in `docs/mcp.md`. **The root identifier does not imply containment**: audience comparison is exact equality, so a root-resource token must not open `/mcp`. Worth a doc comment, because the opposite reading is the confusion attack resource indicators exist to prevent.
- [ ] **`aud` from the bound resource.** `IssueAccessToken` takes the audience per call from `codeData.Resource` / `RefreshTokenData.Resource` — already validated and stored, just never reflected. `VerifyAccessToken` takes the expected audience from its caller. `ValidateResourceIndicator` grows a set. Keep the exactness; a prefix match reintroduces containment.
- [ ] **Verification behind an interface in `internal/auth`**, wired from `main.go` (`internal/api` can't import the oauth package; a verifier *in* `internal/auth` would cycle). The fiddly part is the discriminator: two bearer dialects share one header, so **the issuer decides**, not the verification outcome. `ErrIssuerMismatch` and a malformed JWT mean "not ours, fall through to the provider"; a bad signature, expiry or audience mismatch under *our* `iss` is final — and final means a 401 with `WWW-Authenticate`, never the OIDC provider's `text/html` redirect branch. Table-driven over every error `VerifyAccessToken` can return, with an `oidc`-mode provider behind it.
- [ ] **Identity fidelity — this one is a live bug, not just preparation.** `bearerAuth` builds an identity with no `Email`, but `acl.matchAudience` resolves `user:` against Subject **or** Email. So the grant `docs/authorization.md` uses as its own example, `audience: ["user:*@example.com"]`, matches a browser session and silently fails to match a token from the same person — default-deny, empty cluster, no error. **This affects MCP today.** Add `email`/`name` to `AccessTokenClaims` and reconstruct the full identity; pin the current behaviour with a test first so it lands as a fix.
- [ ] **Operations-tier cap for token callers** (`api.token.operations_level`), mirroring `mcp.operations_level`. It composes with everything — `requireLevel` enforces it, `Allow` reports it, so a client's controls gate themselves. A read-only token is then level 0, which is what a widget or a Shortcut should hold. Reserve `scope` in the claims without implementing it, so adding it later isn't a token-format migration.

### Loose ends from this PR

- [ ] **Enforce the auth invariant in `mcp.New`.** `Handler()` installs `bearerAuth` only `if s.oauth != nil` and otherwise serves mcp-go raw, so `config.ValidateOAuth` states the invariant in a place that cannot enforce it — the only thing between a new caller of `mcp.New` and an unauthenticated `/mcp` is that `main.go` happens to validate first. `New` already returns an error and has `Options.AuthMode` in hand. Blocked on one decision: most existing tests pass no `AuthMode` at all, so a naive `!= "none"` check rejects them.
- [ ] **Lift `mcp.auth_bypass` out of `bearerAuth`** into its own middleware. It currently sits *inside* the middleware that only exists when a verifier does, so an absent authorization server takes the bypass with it — which is why a `cert`-mode deployment cannot run MCP without one, and why `ValidateOAuth`'s error carries a special clause saying so. Lifting it makes that deployment work and deletes the special case.
- [ ] **Retire the `mcp-tokens.json` fallback** once upgrades have had a release to settle. It is the last thing making `internal/oauth` name MCP at all, and with it go the assembled string and the one exemption in `independence_test.go`.
- [ ] Consider whether `oauth.enabled` should default to `true` under a non-`none` auth mode, so the refusal only fires on an explicit contradiction rather than on every existing MCP deployment's first upgrade. Recorded as *considered*: the opt-in was a deliberate call, and the break is documented.

### The client itself

From `docs/specs/2026-09-13-native-apple-client-design.md`. Phase 0 above is the prerequisite; these are sequential.

- [ ] **Decide the type-fidelity approach before any Swift exists.** The Docker types are `type: object` in `api/openapi.yaml` and hand-mirrored in 808 lines of TypeScript, so a generator yields opaque containers for exactly the payloads that matter. Either hand-write Swift `Codable` mirrors plus a fixture-drift test against a live cluster, or tighten the spec and generate both clients — which pays off three times and deletes the existing mirror.
- [ ] **Phase 1 — macOS, read-only (~6–8 weeks).** Multi-cluster sidebar, every resource list and detail, search, history, recommendations, menu bar extra, widgets, Spotlight, Handoff. Topology as a `WKWebView` over the existing React Flow route. TestFlight at the end of this phase, not later. This alone is a product; parity is not a prerequisite and chasing it first is how it never ships.
- [ ] **Phase 2** — logs, metrics in Swift Charts, native topology only if it has earned it.
- [ ] **Phase 3** — writes. Service sub-resources first (that's where the tonnage is), biometric gating on tier-3 operations, App Intents over the same endpoints, optimistic concurrency surfaced as a real conflict UI.
- [ ] **Phase 4 — iOS.** Only once the shared package has been proven by two macOS phases. Live Activities, Controls, StandBy. Decide the push-relay question on evidence about whether anyone wants alerting enough to accept a relay.

### Deliberately not planned

Scopes (the ACL is the authorization model; `Allow` must stay the single answer), flat personal access tokens (strictly worse than a rotating refresh token), cross-accepting MCP and API tokens (the separation is the feature), and a device-management UI — which you *will* want the moment a phone holds a grant, but it is a dashboard page over the existing consent store, not a protocol change.

---

Radiergummi:
Superseded by #248 — same commits on `feat/oauth-authorization-server`, plus the three review fixes. Closing this one; the branch rename is what closed it the first time.

Radiergummi:
## `test` failed on `f22d87a`, and it is not this branch's

```
--- FAIL: TestListStreamEmitsTheDeclaredNames/two_events_inside_the_interval_become_a_batch_array
    asyncapi_frames_test.go:202: no batch frame; got [{Event:service ID:1 …svc1…} {Event:service ID:1 …svc2…}]
```

Two `service` frames where one `batch` was expected — the two events were flushed separately instead of being coalesced.

**Why it is not this branch's.** `internal/api/sse`, `internal/api/asyncapi_frames_test.go` and `internal/cache` are byte-identical between `main` and this head — `git diff origin/main f22d87a --` over those three paths is empty. What this branch changes under `internal/api` is three comments, four identifier spellings in `router.go`, and one renamed test fixture; nothing reaches the broadcaster, the batching loop, or the test. The test also passes here on this head, 20 runs in a row.

**What is actually racy.** The subtest builds its router at a 50 ms batch interval and broadcasts twice back-to-back inside a 400 ms stream. Between the two `Broadcast` calls there are three goroutine hops — `Broadcast` → `inbox` → `fanOut` → the per-client `events` channel → the streaming loop — and that loop selects between `client.events` and `batchTicker.C`. If a tick becomes ready while the first event is being appended, the tick can win the select even though the second event is already queued, because `select` chooses uniformly at random among ready cases. The first event is then written alone as `event: service`, and the second follows in the next window. Nothing about the flush is wrong; the test just assumes an ordering the loop does not promise. It is rare — it survived 400 targeted runs and four whole-package runs here under 8–12× CPU oversubscription — which fits a loaded runner hitting it once.

**No fix exists to port**, so this is a proposal rather than a push; widening an OAuth extraction with an unrelated test change is worse than leaving it. The smallest fix is to stop involving the ticker at all: give that subtest a batch interval longer than the stream's own lifetime, so both events are still pending when the context is cancelled and the context-done path flushes them as one batch.

```diff
-		router, _, broadcaster := resourceStreamRouter(t)
+		router, _, broadcaster := streamTestRouter(t, time.Second)
```

That makes the assertion deterministic — the batch is flushed by cancellation, not by a tick that may or may not have fired — and still tests the claim the subtest is named for, that two events inside one interval arrive as an array. Verified here: 20 consecutive runs green with the frame arriving as `event: batch` carrying two envelopes.

Re-running the failed jobs once, as the one re-run this warrants.

---

PR #246, opened 2026-09-13
feat: mint API access tokens, and accept them as bearer credentials

Radiergummi:
Phase 0's second half, on top of #244. The authorization server now issues tokens for the REST API as a second protected resource, and the auth middleware accepts them — so a script, a CLI or a native client can authenticate unattended, on any auth mode, with no operator configuration beyond `oauth.enabled`.

**Stacked on #244.** Base is `claude/native-macos-ios-swarm-app-gqagnx`, so this diff is the commits below rather than the extraction underneath them.

Implements decisions 1–4 of `docs/specs/2026-09-13-api-oauth-tokens-design.md`. Decision 5 (the operations-tier cap) is deferred — its premise was wrong, see below.

All 11 checks green on `e83f012`.

## Commits, in the order they want reading

| | |
|---|---|
| `b01b47c` | **fix(oauth):** carry the provider's identity into the token — a live bug, independent of the rest |
| `16fb885` | **feat(oauth)!:** issue tokens for more than one protected resource |
| `ab55d35` | **feat(auth):** accept API bearer tokens in the middleware |
| `5f28c07` | **docs:** the user-facing story, and a catalog link the resource move broke |
| `8a7d158` | **refactor(oauth):** settle the resource set at construction — quality pass, no behaviour change |
| `2d3f313` | **fix(oauth):** refuse to found a grant on a token this server issued — **security** |
| `9309cf2` | **fix(oauth):** serve discovery where the RFCs say a client will look — **conformance** |
| `1baa3b5` | **fix(oauth):** close the rest of the conformance gaps |
| `21b8b60` | **test(oauth):** name fixtures so gosec does not read them as credentials |
| `e83f012` | **feat(oauth):** say there are no scopes, rather than saying nothing — **conformance** |
| `5ba2d6f` | **docs(spec):** record the one scope the no-scopes reasoning does not cover |

## The live bug, first

`acl.matchAudience` resolves a `user:` audience against Subject **or** Email, but a token carried neither beyond the subject. So the grant `docs/authorization.md` uses as its own example —

```yaml
- resources: ["stack:public-*"]
  audience: ["user:*@example.com"]
  permissions: ["read"]
```

— matched a browser session and silently missed a token from the same person. The ACL is default-deny, so the client saw an empty cluster, no error, and nothing to explain it. **This affects MCP as shipped**, not just the new API tokens.

It ran deeper than the claims: `AuthCodeData` and `RefreshTokenData` did not carry the fields either, so the fix runs the full length of the flow — consent, the code, the access token, the refresh token, and the persisted snapshot. The test drives consent and the code grant rather than handing `IssueAccessToken` a fixture, because the defect was in the propagation and a fixture would have proved the claim round-trips while the flow still dropped it.

## Two resources

A resource is a path under the deployment root; the identifier and the RFC 9728 document both derive from it, and `aud` comes from the resource the grant was bound to rather than from a constant. Comparison is exact per element — a prefix match would make a token for the root reach everything beneath it, which is the confusion the resource indicator exists to prevent.

| Resource | Identifier | Metadata document |
|---|---|---|
| REST API | the deployment root | `/.well-known/oauth-protected-resource` |
| `/mcp` | the deployment root + `/mcp` | `/.well-known/oauth-protected-resource/mcp` |

`oauth.api_tokens` (default on) offers the API, and it comes first so a client sending no indicator gets a token for the deployment it pointed at. With it off, `/mcp` is both the only resource and the default — exactly today's behaviour.

**Wire-visible:** `/mcp`'s metadata moves off the root location, because one document can describe only one resource and the root now belongs to the API. A client that follows `resource_metadata` from the 401 is unaffected; one that hardcoded the root path breaks. Changelog and `docs/mcp.md` both updated.

## Security: an access token could found a new grant

`2d3f313`. A review found the design doc's own assumption broken by this branch:

> `/oauth/authorize` stays non-exempt too — consent must run under a real session, which is precisely how the upstream provider's identity gets into the token.

"Non-exempt" meant "the provider authenticates it" when written. Once a token of ours satisfied the middleware first, it stopped meaning that.

Not a wider identity — claims travel verbatim. It is **persistence and laundering**: a holder of one leaked 1-hour token could register their own client (DCR is unauthenticated by design), drive consent with the stolen token, and redeem a 30-day rotating refresh token in a grant family under a `client_id` the user never saw. Revoking the original grant does not touch it, theft detection never fires, and no consent record is written for an unverified client. It also crosses the API → `/mcp` audience boundary, in the direction on by default.

Worst under `cert`, `headers` and `tailscale`, where the provider's credential is transport-bound and could not previously be replayed there at all. Consent now refuses an identity whose provider is the authorization server itself, keying on `Identity.Provider` — what that field was added for, and which nothing previously read. Both verbs checked independently; without the fix the POST answers `302` with a code.

## Standards conformance

Two passes. The first three, in `9309cf2`:

- **RFC 9728 §3.1** — the metadata URL was wrong under a base path, and an earlier commit here *defended* that as deliberate. The defence was about reachability and the question was conformance: a client **derives** the URL rather than reading ours. Both locations are served now; the challenge still advertises the mounted one, because that URL is followed rather than derived and must resolve through whatever proxy exists. What to forward is documented rather than assumed away. Reaching the derived URL also needed `basePathMiddleware` to stop 404ing paths without the prefix.
- **RFC 8707 §2** — a repeated `resource` parameter was silently narrowed to the first, handing back an audience the client had not settled on. Now `invalid_target`.
- **RFC 6750 §3.1** — a 401 carrying no credential reported `invalid_token`. `/mcp` sent that on every client's opening request.

Eight more in `1baa3b5`, of which three matter most:

- **RFC 9110 §15.5.2** — the consent refusal added above emitted a 401 with no challenge, and used 401 for two different answers. The token-provenance refusal is now **403**: the request was authenticated, and resending the same kind of credential cannot help.
- **RFC 7515 §4.1.9 / RFC 2045** — `typ` is a media type, so its case is insignificant. Compared byte-exact, so `AT+JWT` was refused as malformed → classified as another issuer's → handed to the upstream IdP.
- **RFC 7519 §4.1.3** — `aud` is an array in the general case. Typed `string`, so the one-element-array form most libraries emit took that same misrouting path.

Plus `nbf` (RFC 7519 §4.1.5), required claims at verify time, the RFC 8414 §3 equivalent for AS metadata (and deliberately *not* for `openid-configuration`, whose rule genuinely differs), RFC 9207's `authorization_response_iss_parameter_supported` (without which our `iss` mix-up defence was unenforceable), RFC 3986 §6.2.3 trailing-slash equivalence on the root identifier, and RFC 6749 §5.2 error codes — an absent parameter answered `invalid_grant`, which tells a client to discard a working refresh token.

And one more in `e83f012`:

- **RFC 8414 §2** — `scopes_supported` is RECOMMENDED and both documents omitted it entirely, which a client reads as *unspecified* rather than *none*, leaving it to guess at a scope this server would ignore. Both now carry `"scopes_supported": []` — present, not omitted, because that distinction is the point of the field.

Scopes themselves stay out, deliberately: nothing is gated on one, so `offline_access` or a per-transport scope would advertise an authorization boundary that does not exist. What that leaves is the conformant handling of having none — a requested scope is ignored rather than refused with `invalid_scope` (refusing would break clients that send `offline_access` unprompted, and the grant is unaffected either way), and the token response omits `scope`. **This supersedes an earlier note here calling RFC 6749 §5.1 a gap: it is not.** §5.1 requires the parameter only when the granted scope differs from the requested one, and with no scopes defined both reduce to the empty set. Echoing `scope: ""` would violate the parameter's own ABNF.

## Three corrections to the design doc

1. **`ErrIssuerMismatch` was unreachable.** The signature was verified before the payload decoded, so a token from another issuer reported `ErrInvalidSig` and would have been refused finally instead of passed on. The issuer check runs first now.
2. **The OIDC redirect trap was already closed** — the provider checks the bearer header before its `text/html` branch.
3. **Decision 5's premise does not hold.** `requireLevel(required, configured)` decides at **construction** time and `h.operationsLevel` is a struct field, so a per-caller tier cap needs a request-scoped level across three constructors and three `allow.go` reads. Deferred rather than bundled.

`5ba2d6f` adds a fourth note under that decision rather than a correction: a scope naming the operations tier is the one scope proposal the no-scopes reasoning does not cover, because it only ever subtracts — `min(deployment tier, transport cap, requested)`. It is the same cap wearing a scope as its front end, so it is gated on the same request-scoped level and must follow it rather than lead; a scope advertised but enforced nowhere is exactly what was rejected. Recorded so the question does not read as settled.

Also settled, unasked by the design: a request carrying both a cookie and a token. **Token wins** — an explicit credential outranks an ambient one, and the alternative silently ignores a token the client deliberately sent.

## Verification

All 11 checks green. Locally: `golangci-lint run` (15 linters) 0 issues, `golangci-lint fmt --diff` clean, `go build`, `go vet`, `go test ./...`, frontend `oxlint` + `tsc`, website `oxlint` + `astro check`.

The release `golangci-lint` binary refuses this repo's Go version (built with 1.25, targets 1.26), which is how two gosec findings reached CI. Building v2.12.2 from source with the local toolchain resolves it and reproduces the gate exactly:

```
go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2
```

Beyond tests, the built binary was driven directly. `/profile` over a real API token returns `email` and `displayName` that would both have been empty before `b01b47c`. The discriminator, live: an `/mcp`-audience token gets a 401 naming the **API's** document; a wrong-key token gets a final 401; a token under another `iss` falls through to the provider, whose 401 carries no `resource_metadata`. And on a base-path deployment, both metadata spellings return the same document — the derived one 404s with the middleware change backed out.

## Noted, not fixed

- **Every `/oauth/*` endpoint answers `503 ENG001` while Docker is unreachable.** `readsClusterState` exempts `/.well-known/` but not `/oauth/`, so DCR, authorize and token fail even though none reads cluster state. Pre-existing and affects MCP on `main` the same way, but it does mean a client cannot obtain a token while the cluster is down.
- `internal/auth/cert.go` advertises `mutual-tls`, which is not a registered scheme — RFC 8120 registers `Mutual`. Pre-existing.
- `httpQuotedString` passes control characters through where `qdtext` excludes them. Inert: Go's header writer neuters CR/LF and both current inputs are hardcoded.

https://claude.ai/code/session_016xKFmYiLH4EcuukDgY1pfn


Radiergummi:
## Two review passes since the description was written — one of them retracts part of it

### Retraction: the RFC 9728 deviation was wrong

The description, under **Wire-visible**, defends this:

> One deliberate deviation: RFC 9728 §3.1 inserts the well-known segment after the *authority* … The base path therefore precedes the well-known segment, matching where the AS metadata has always been served. Stated in the doc comment so it reads as a decision rather than an oversight.

Treat that paragraph as withdrawn. The reasoning was about *reachability* and it settled a question about *conformance*, which is not the same thing — a client does not read our location, it **derives** it from the resource identifier. So under a base path a conformant client asked for a URL we did not serve, and the doc comment made an oversight read as a decision. The MCP ecosystem enforces §3.1 directly; [SEP-985](https://modelcontextprotocol.io/seps/985-align-oauth-20-protected-resource-metadata-with-rf) exists for it.

Fixed in `9309cf2`. Both locations are served: the §3.1-derived one (well-known after the host, base path inside the path) and the one beneath `server.base_path`, which stays because a proxy forwarding only `{base_path}/*` is the reason the deviation was tempting and an operator cannot always change it. They are the same address when no base path is set. The challenge still advertises the mounted spelling — that URL is *followed* rather than derived, so it must resolve through whatever proxy already exists — and what to forward is now documented in `docs/mcp.md` instead of assumed away.

Reaching the derived URL also needed `basePathMiddleware` to stop 404ing paths without the prefix, which had made the conformant location unreachable however the routes were registered. `/.well-known/` now passes through unstripped.

Two smaller conformance fixes rode along:

- **RFC 8707 §2** — the `resource` parameter may repeat, to request a token valid at several resources. Every call site read only the first and dropped the rest, so the client got an audience it had not settled on: the confusion resource indicators exist to prevent, reached by our own hand. Now refused with `invalid_target`.
- **RFC 6750 §3.1** — a 401 for a request carrying *no* credential should not report an error code. `/mcp` sent `error="invalid_token"` on every unauthenticated first request, which is the opening move of every MCP client handshake. Now omitted when there was no credential, kept when one failed.

### Security: an access token could found a new grant

`2d3f313`. A security review found that the design doc's own assumption had been broken by this branch:

> `/oauth/authorize` stays non-exempt too — consent must run under a real session, which is precisely how the upstream provider's identity gets into the token.

"Non-exempt" meant "the provider authenticates it" when that was written. Once a token of ours satisfied the middleware first, it stopped meaning that, and the consent endpoint began accepting the credential it exists to mint.

Not a wider identity — claims travel verbatim, so the ACL still binds a minted token to the same person. It is **persistence and laundering**: a holder of one leaked 1-hour token could register their own client (DCR is unauthenticated by design), drive consent with the stolen token, and redeem a 30-day rotating refresh token in a new grant family under a `client_id` the user never saw. Revoking the original grant does not touch it, theft detection never fires, and no consent record is written for an unverified client — so there is neither a remedy short of rotating the signing key nor a signal that one is needed. It also crosses the API → `/mcp` audience boundary this series exists to enforce, in the direction that is on by default.

Worst in `cert`, `headers` and `tailscale`, where the provider's credential is transport-bound and previously could not be replayed here at all. Under `oidc` a stolen IdP token could already drive consent, but that is full account compromise by then.

Consent now refuses an identity whose provider is the authorization server itself — keying on `Identity.Provider`, which is what that field was added for and which nothing previously read. Both verbs are checked independently; without the fix the POST answers `302` with an authorization code.

### Still open

- **CI has never run on this PR.** `ci.yml` triggers on `pull_request: branches: [main]`, and this targets #244's branch — so no `pull_request` event matches. It should start once #244 merges and GitHub retargets this to `main`. Consequently the 15 linters in `.golangci.yml` (`gosec`, `gocritic`, `modernize`, `errorlint`, `musttag`, …) have not run; `golangci-lint` refuses this repo's Go version locally, so only `golines` was verified standalone. `go build`, `go vet`, `gofmt` and `go test ./...` are clean.
- **RFC 8414 §3 has the same insertion rule** for `/.well-known/oauth-authorization-server`, and that document is still only at the mounted location under a base path. Pre-existing rather than introduced here, but it is the same class and now a two-line change given the mechanism above. Happy to take it.
- **`ExtractBearerToken`** (`internal/auth/oidc.go`, pre-existing) slices `auth[7:]`, so `Bearer‚ÄÖ‚ÄÖtoken` with two spaces yields a token with a leading space, where RFC 9110's `credentials` ABNF allows `1*SP`. It does not validate the token68 charset either. My new path depends on it.
- **`nbf` is not validated** in `VerifyAccessToken`. RFC 7519 §4.1.5 says a JWT must not be accepted before its `nbf`. We never mint one, so nothing is exploitable — pure conformance.

---

Radiergummi:
Superseded by #251 — this PR closed automatically when the branch was renamed from `claude/bearer-token-api-auth-8f185r` to `feat/bearer-token-api-auth`, and a PR's head branch cannot be moved. Same commits, plus four review fixes. The discussion above stays here.

PR #248, opened 2026-09-13
refactor(oauth)!: extract the authorization server, and make it opt-in

Radiergummi:
> Supersedes #244, which GitHub closed when its branch was renamed. Same commits, plus three
> review fixes on top: refresh-token theft detection reaching `Rotate` again when the client
> sends `resource`, the dev auth stack's trusted-proxies variable, and the stale doc rows that
> still promised a one-time `mcp-tokens.json` fallback — no fallback shipped, so the note about
> the state-file migration below no longer holds.

Two design investigations, then the first piece of work they call for.

## Breaking changes

| was | now |
|---|---|
| implicit (MCP + any auth mode but `none`) | `oauth.enabled` / `CETACEAN_OAUTH_ENABLED`, default **false** |
| `mcp.issuer`, `mcp.signing_key`, `mcp.access_token_ttl`, `mcp.refresh_token_ttl`, `mcp.consent_ttl` | `oauth.*` / `CETACEAN_OAUTH_*` |
| `[mcp.oauth]` (`require_resource_indicator`, `dcr_*`, `cimd_enabled`) | `[oauth]` |
| `mcp.oauth.auth_bypass` | `mcp.auth_bypass` (stays MCP's) |
| `{data_dir}/mcp-tokens.json` | `oauth-tokens.json`; the old name is not read, so delete it |

Access tokens issued before this stop verifying and the published `kid` changes, because the key-derivation labels moved too. Refresh tokens do not survive either: `readState` accepts only version 2 and no fallback shipped, so every client authorizes once more.

## Why

`docs/specs/2026-09-13-api-oauth-tokens-design.md` plans to make the server issue tokens for the REST API as a second protected resource, which needs it shared rather than MCP's. Extracted first, the diff that adds a second audience is reviewable; done together it would be buried in a rename.

It also becomes opt-in, which was the repo owner's call and turned out to simplify two things rather than complicate them. `OAuthIssuerRequired` used to approximate "will anything drive the authorize flow" from the auth mode and the bypass list; with an explicit flag it *is* the flag, and `internal/config`'s OAuth half now names MCP nowhere. And the opt-in decouples the server's lifecycle from `mcp.enabled` by itself, which the design doc had deferred to a later change.

## The part worth reviewing carefully

**Three combinations are settled at startup** rather than discovered later. MCP under any auth mode but `none` *without* the server is refused, because `/mcp` is exempt from the auth middleware and authenticates itself — with no server there is no bearer check, and the endpoint would serve the cluster's write surface unauthenticated. `mcp.auth_bypass` does not excuse it, and the error says so: the bypass runs *inside* the very middleware an absent server never installs. The server under `none` is refused too (consent would ask an anonymous user to approve a client). The server with nothing consuming it is allowed and warned about.

**The nil-interface guard in `main.go`.** `Options.OAuth` became an interface, and `oauthSrv` is nil when no server was built. A nil pointer inside an interface is not nil, so the existing check in `Handler()` would have passed, `bearerAuth` would have armed, and every unauthenticated `/mcp` request would have panicked on a nil receiver. `govet`'s nilness does not model typed-nil-in-interface and `main.go` has no tests, so the guard follows the precedent the Prometheus client already set six lines above.

**Both golden constants were recomputed from the curve parameters**, not copied from the failing test output — `TestGoldenKID` exists to catch drift in go-jose's JWK encoding, and pasting whatever the code now produces would make it a tautology. The method was validated by reproducing the two committed values first.

## Shape of the series

Eight commits, each green. `git mv` is content-free so all 27 files report `R100` and `git log --follow` still works across the boundary; the renames land after, on files git has already learned the new path for. The first commit only pins behaviour the later ones change — the identity a bearer token yields, the `WWW-Authenticate` value, the CSRF cookie's wire name — so each effect shows up as a one-line diff rather than an absence.

`internal/mcp`'s *production* code no longer imports the package; its tests still do, deliberately, because they mint real ES256 tokens and are the only proof verification works end to end.

`internal/oauth/independence_test.go` now enforces the point of the exercise by scanning the package's own source for its consumer's name. It carries no exceptions.

## Verification

`make test` and `make typecheck` are green (the local `golangci-lint` is built against Go 1.25 and refuses a 1.26.6 target, so CI owns `lint`/`fmt-check`). Live checks: both discovery documents serve with the right issuer and resource, and all three validation refusals behave. `/oauth/jwks` and `/mcp` could not be exercised here — they sit behind a Docker-reachability gate and this container has no daemon — so the `kid` change and the unguarded-`/mcp` path rest on `TestGoldenKID` and `TestHandlerWithoutOAuthServesUnguarded`.

## Also on this branch

The two investigations that prompted it: `2026-09-13-native-apple-client-design.md` (a native macOS/iOS client — feasible, with credentials as the blocking gap and multi-cluster as the case the web dashboard structurally cannot make) and `2026-09-13-api-oauth-tokens-design.md` (Phase 0 in detail: resource identifiers, the RFC 9728 metadata layout, and where token verification runs). Decisions 6 and 7 of the latter are annotated where this work went the other way.

https://claude.ai/code/session_01BByYVLBwsLhty9aFdoSPUy

---

## Review fixes (`a96d845a`)

Three findings from a review of this branch, fixed on it:

- **`/mcp` answered 401 with no `WWW-Authenticate`** (`internal/mcp/server.go`). RFC 9110 §15.5.2 makes the header a MUST, and the guard's own comment said it had none to offer — which is this release's own argument for answering `403` under `cert`, `tailscale` and `headers`. Now 403, matching `originGuard` beside it and the API endpoints.
- **The consent CSRF token covered five of the form's eight hidden fields** (`internal/oauth/consent.go`). `response_type`, `resource` and `code_challenge_method` were unbound. Inert today, but the second protected resource this extraction exists for makes `resource` live.
- **Protected resource metadata named `/mcp` even with `mcp.enabled` false** (`main.go`, `internal/oauth/prm.go`), sending RFC 9728 discovery to a path that 404s. A `ResourceMounted` flag makes the document absent rather than wrong, as `catalogMounts` already did for its link.

`go test ./...`, `go vet -tags e2e ./...` and `golangci-lint run` are clean. No changelog entry: the 403 rule is already written down, and the other two correct behaviour this release has not shipped.


Radiergummi:
Carrying one finding over from #244 so it isn't buried in a closed PR — nothing here blocks this one, CI is green on `f22d87a`.

The first suite on this SHA failed `TestListStreamEmitsTheDeclaredNames/two_events_inside_the_interval_become_a_batch_array`; the `test` job has since passed twice more on the same commit, so it was a flake, and the test is pre-existing on `main` rather than anything this branch touches. It is genuinely racy: the subtest broadcasts twice at a 50 ms batch interval, and the streaming loop selects between `client.events` and `batchTicker.C` — a tick that becomes ready while the first event is being appended can win the select even with the second already queued, since `select` randomizes among ready cases, so the first event flushes alone as `event: service`.

[The #244 comment](https://github.com/Radiergummi/cetacean/pull/244#issuecomment-5655964248) has the full reasoning and a one-line fix that removes the ticker from the assertion (give that subtest a batch interval longer than the stream's lifetime, so cancellation flushes both events as one batch — 20 consecutive green runs locally). Worth its own change rather than this one.

---

PR #266, opened 2026-09-14
Draw Traefik routing and stack contents as graphs

Radiergummi:
Two graph views, and the shared shell underneath them.

## Traefik routing, on a service

The Traefik panel rendered routers, services and middlewares as three columns of boxes — what the labels *contain*, but not what they *mean*. A router's middleware chain is ordered, its service binding is often implicit, and a reference like `auth@file` points somewhere those labels cannot see; none of that survives a list.

The panel gains a **Graph** view beside Structured and Labels, drawing entrypoints into routers, through the middleware chain in order, to the services they reach. References the labels do not define become nodes marked external rather than absent arrows — a missing arrow reads as "no middleware" instead of "defined elsewhere".

## Stack contents, on a stack

A stack page listed its services, configs, secrets, networks and volumes in five tables. Which services share an overlay, which mount the same secret, and which resources nothing uses at all are answers those tables hold but do not show.

The page now opens with **Topology**: networks on the left, services in the middle, the configs, secrets and volumes they mount on the right, every node a link to the resource it stands for. A service reads its running tasks against the number wanted, from the poll the table beside it already runs. It does not duplicate the Topology page, which draws networks as *edges* between services and omits configs, secrets and volumes entirely.

Two details worth knowing, both tested: volume mounts match on `Source` name rather than an ID, because that is how volumes are keyed here; and bind mounts name a host path, so they are skipped.

## Layout

ELK, from sizes React Flow measured rather than guessed. Laying out from constants put edges in the empty half of an oversized box, and hand-rolled column placement could not keep an edge that skips a column off the nodes between it. ELK's bend points are kept and drawn as one smooth spline. Entrypoints and networks are pinned to the first layer, services and mounts to the last, so the reading still runs left to right.

React Flow and ELK stay off the page's own chunk: each graph is `lazy()`, and its frame defers mounting until scrolled into view, inside a section that renders nothing while collapsed. Verified against a production build — `vendor-elk` is not preloaded and is reachable only through a dynamic import.

Both graphs share a bounded viewport: panning clamps to the laid-out bounds plus a margin, zoom cannot fall far below the point where the whole graph is on screen (a floor taken from the first fit, since a useful minimum depends on the graph), and zoom, reset and full screen sit in the corner, easing between viewports rather than jumping. Viewport animation follows `prefers-reduced-motion`, dropping the movement rather than the move.

## Reading the graph

React Flow ships wired for an editor, and turning that off is most of the accessibility work here. By default every node *and every edge* is a tab stop — twenty-three on the Traefik graph before the controls — each announced through the same canned description: press enter to select, press delete to remove. An edge read out the ids of its ends, which are Docker's. None of it is true of a read-only graph.

Nodes and edges are no longer focusable, and the canvas carries a name instead of a bare `role="application"`. The Traefik nodes held everything they knew in a tooltip a pointer had to hover to open; they are buttons now, so the rule, the priority, the resolver and the TLS domains are reachable by tab.

Hovering or focusing a node dims everything more than one edge from it. Eleven edges crossing is what the layout costs on a graph worth drawing, and following one across is work the reader should not have to do. Focus is the keyboard's hover, so both drive it; only focus is addressable, so only focus names the node in the stack page's URL (`?node=…`), and a link arriving with one centres on it once the fit has settled.

## Checks

`make check` equivalents all clean — oxfmt, oxlint, `tsc -b`, 814 frontend tests, `go build ./...`. The transforms are tested first and in full; the layout itself is ELK's.

Because jsdom performs no layout, the visual invariants are pinned by Playwright instead: every edge is sampled along its path and compared against every node box and every other edge, so a line drawn through a label or across another fails rather than merely looking wrong. Exact positions are left alone — that is ELK's business, and pinning it would break on a version bump. The Traefik graph serves a fixed integration from a route, shaped around what the layout has to get right: a long chain, an edge that skips it, an undefined reference, an unused middleware, and a router whose service cannot be resolved.

Two gaps, both stated in the code: deep-link centring is not covered, because jsdom reports zero-size nodes and the effect returns early; and `edgesFocusable={false}` is not covered, because with no layout React Flow renders no edges for a selector to match.

https://claude.ai/code/session_018i2rhcGti9MCqECRpJavYb


Radiergummi:
Added the e2e coverage the description called out as missing — `3eb2d35c`.

`frontend/e2e/graphs.spec.ts`, 10 tests across both graphs. Every edge is sampled along its path and compared against every node box and every other edge, so a line drawn through a label, or across another line, fails rather than merely looking wrong. Plus: the whole graph lands inside its frame, every stack node is a link, a link opens the resource it names, the controls exist, zoom will not fall past the fit, and reset recovers.

Exact node positions are deliberately not asserted — that is ELK's business and pinning it would break on a version bump.

**The Traefik fixture is served from a route.** The cluster a run points at is not ours to arrange, and the graph is drawn from the parsed integration alone, so `stubTraefikLabels` injects a fixed one shaped around the cases that actually broke: a chain several steps long, an edge skipping those steps, an `auth@file` the labels do not define, an orphan middleware, and a router whose service cannot be resolved. This means the test cannot catch a regression in how the *server* parses labels — `internal/integrations` is what covers that. The stack graph needs no fixture; every swarm has one.

Two things worth knowing for running them:

- They need a Cetacean serving *this* frontend. A deployed instance on the default `:9000` may be an older build, where they time out waiting for a canvas that isn't there. There's a comment at the top of the spec saying so.
- Against a Vite dev server with parallel workers they are flaky — the dev server times out compiling lazy chunks on demand, not the tests. Against a built binary: 10/10 across seven consecutive runs, ~5s.

I mutation-checked the two that matter rather than trusting green: dropping the routed edge so lines run straight through nodes fails the crossing test, and turning the stack nodes back into plain divs fails the link test. Both pass again on restore.

https://claude.ai/code/session_018i2rhcGti9MCqECRpJavYb

docs/integrations.md
```diff
@@ -29,12 +29,29 @@ A panel shows the tool as enabled unless a `<prefix>enable` label says otherwise
...
+
+The graph is drawn from this one service's labels. A middleware or service the labels reference but do not define,
+such as `auth@file`, appears dashed and marked as defined elsewhere — Cetacean cannot see the configuration behind
+it. A router naming no service binds to the only service the labels declare, and is drawn unresolved when there is
+none or several. Middlewares and services nothing references stay visible, dimmed.
+
+The graph pans and zooms within its own bounds, so it cannot be pushed out of view, and carries controls for zoom,
+fit and full screen.
+
+For the configuration Traefik is actually running, across every service in the cluster, use Traefik's own dashboard.
```
Radiergummi: This is *way* too wordy for a small visualisation.

PR #280, opened 2026-09-16
fix(cluster): resolve a task name to the live replica in the slot

### LINKED ISSUE #227: mcp: `resolveTask` resolves `<service>.<slot>` to a replaced task, not the live one
## Summary

`internal/mcp`'s `resolveTask` resolves a task addressed as `<service>.<slot>` to the **first** name match in `(slot, ID)` order rather than to the task actually running in that slot. Swarm keeps a record for every replica it has replaced, so any service that has been updated — or that restarts in a loop — has several tasks in one slot, and this returns whichever has the lowest ID. That is usually a terminal record.

So `describe`, `get_logs` and `remove_task` on `web.3` can report a failed historical replica as though it were the current one. Deterministic, and deterministically wrong.

## Why it happens

`Cache.ListTasksByService` sorts by `(Slot, ID)` (`internal/cache/cache.go:759`), and `resolveTask` returns the first task whose `cluster.TaskName(task, &service)` equals the identifier (`internal/mcp/resolve.go:50-55`). `TaskName` renders every task in slot 3 as `<service>.3`, so the scan cannot tell a live replica from a replaced one and takes whichever it reaches first.

## Reproduction

A cache holding one service and two records for slot 1 — a terminal one whose ID sorts first, and the live replica that succeeded it:

```go
c.SetTask(swarm.Task{
    ID: "aaa-dead", ServiceID: "svc1", Slot: 1,
    DesiredState: swarm.TaskStateShutdown,
    Status:       swarm.TaskStatus{State: swarm.TaskStateFailed},
})
c.SetTask(swarm.Task{
    ID: "zzz-live", ServiceID: "svc1", Slot: 1,
    DesiredState: swarm.TaskStateRunning,
    Status:       swarm.TaskStatus{State: swarm.TaskStateRunning},
})

task, _, _ := srv.resolveTask("web.1")
```

Observed:

```
web.1 resolved to "aaa-dead" (desired=shutdown, state=failed)
```

Expected `zzz-live`.

## Proposed fix

Prefer the live task in the slot, where live is `cache.TaskIsLive` (`DesiredState` is neither `shutdown` nor `remove`) — already the shared predicate behind every replica count and both placement views. A slot holding only terminal records should resolve to nothing rather than to the most recent of them: a caller addressing `tasks/3` means the replica running there, and there isn't one.

The fix should not be applied in `internal/mcp` alone. REST is gaining name-addressed URLs (`internal/api/canonical.go`), and a task identifier that means one thing over MCP and another over REST is the disagreement that work exists to have ended. Task resolution should move to `internal/cluster` as `ResolveTask(c *cache.Cache, identifier string)` — which is where `CLAUDE.md`'s own note already puts it, since the name is derived by `cluster.TaskName` and the
cache cannot own a rule that depends on a package importing it — with `internal/mcp`'s `resolveTask` delegating to it.

## Notes

Found while writing `docs/specs/2026-09-11-friendly-urls-design.md`, whose "Addressing a task" section specifies the rule and its shared home. Global services are unaffected by the slot ambiguity: their tasks render as `<service>.<node>`, which is unique per task.

Radiergummi:
Closes #227.

`resolveTask` resolved `<service>.<slot>` to the **first** name match in `(Slot, ID)` order. Swarm keeps a record for every replica it has replaced, so any service that has been updated — or that restarts in a loop — has several tasks rendering under one name, and the lowest ID wins. That is usually a terminal record, so `describe`, `get_logs` and `remove_task` on `web.3` could report a failed historical replica as the live one.

## The rule

Among the tasks rendering under an identifier, the live one wins — live being `cache.TaskIsLive`, the predicate already behind every replica count and both placement views. A slot holding only terminal records resolves to nothing rather than to the most recent of them: a caller naming a slot means the replica running there, and there isn't one.

Addressing a task by its own ID is untouched, which is how the task detail view and `get_logs` keep reading history.

## Where it lives

Resolution moves out of `internal/mcp` into `cluster.ResolveTask`, with `resolveTask` delegating. REST is gaining name-addressed URLs and derives the name from `cluster.TaskName`, so fixing this in place would leave the two transports disagreeing about what `web.3` denotes — the disagreement `internal/api/canonical.go` exists to have ended. The cache cannot own the rule, since the name is derived by a package importing it.

`canonical.go`'s note that tasks are excluded because no resolver exists is updated: one exists now, and tasks stay out of the flat name form because their names are service-scoped rather than unique.

## One correction to the issue

The issue says global services are unaffected because `<service>.<node>` is unique per task. They are not. Updating a global service leaves the replaced record on the node now holding its successor, and both render identically — `TestResolveTaskPrefersTheLiveTaskOnANode` returned `aaa-dead` before the fix, exactly as the slot case did. So the preference applies uniformly rather than to replicated mode alone, and the "Addressing a task" section of `docs/specs/2026-09-11-friendly-urls-design.md` now says so outright instead of by omission.

## Tests

Nine in `internal/cluster/resolve_test.go`, one in `internal/mcp/resolve_test.go` pinning that the MCP path is the one applying the rule. The logic was moved verbatim first, without the preference, so the three new-behaviour tests were watched failing against real code:

```
resolve_test.go:37:  ResolveTask(web.1) = "aaa-dead", want "zzz-live"
resolve_test.go:67:  ResolveTask(web.1) = "old", want no task
resolve_test.go:233: ResolveTask(agent.node-2) = "aaa-dead", want "zzz-live"
```

The other six cover behaviour the move had to preserve: dotted service names, unplaced globals, a terminal task by ID, ambiguity errors, and a plain miss.

https://claude.ai/code/session_01EBGX2tEBaVmR3ANHNQuYkN


Radiergummi (commented):
lgtm

